PR #183, opened 2015-03-27
Extend highlights slightly past the first and last moves of a line.

lgarron:
This was really bothering me. The lines still don't extend to the end of the cell, but it's way better.
In most cases, the font doesn't even get smaller.

0.8.12:
screenshot 2015-03-27 06 51 51

After this change:
screenshot 2015-03-27 06 51 47

### DIFF
--- a/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
+++ b/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
@@ -91,7 +91,7 @@ class ScrambleRequest {
     private static final int MIN_LINES_TO_ALTERNATE_HIGHLIGHTING = 4;
     private static final BaseColor HIGHLIGHT_COLOR = new BaseColor(230, 230, 230);
     private static final int SCRAMBLE_PADDING_VERTICAL = 3;
-    private static final int SCRAMBLE_PADDING_HORIZONTAL = 3;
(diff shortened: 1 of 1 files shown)


jfly:
Clever hack =). Just so I have something to link to if people ever complain:

Our left and right horizontal padding is now a good bit larger than it used to be. Are we ok with that tradeoff?

Have you checked every puzzle to make sure that this sudden loss of horizontal space doesn't have an unexpected fallout?

jfly:
We just merged in some more cosmetic spacing changes from Pedro here: https://github.com/cubing/tnoodle/pull/184. This stuff is all kind of guess and check, so please pull down the latest changes and check how they interact with yours.

If all is well, we can merge this up today, make a release, and push it out on Monday.

jfly:
Oh, and in the future, can you do this in a fork? I don't want to dirty up the cubing/tnoodle branch.

lgarron:
> Our left and right horizontal padding is now a good bit larger than it used to be. Are we ok with that tradeoff?

Yep.

> Have you checked every puzzle to make sure that this sudden loss of horizontal space doesn't have any unexpected fallout?

Yep.

> We just merged in some more cosmetic spacing changes from Pedro here: #184. This stuff is all kind of guess and check, so please pull down the latest changes and check how they interact with yours.

Pedro just sent an email. Square-1 is too vertically cozy, but I still think it's better than before.

> Oh, and in the future, can you do this in a fork? I don't want to dirty up the cubing/tnoodle branch.

Sure.

PR #184, opened 2015-03-27
Bigger leading -  more space between lines

pedrosino:
I've changed the cell leading to 1.1*fontSize, to make the text more readable.

It still fits inside the cell, because there was space left at the top and bottom.


jfly:
A bit scary to me that we don't know how to compute the amount of space these things take up, but whatever. If it works, it works.

Thanks, Pedro!

lgarron:
Can't we add 0.1 * (# lines - 1)?

pedrosino:
Where exactly?

PR #185, opened 2015-03-27
Change "may not" to "must not".

lgarron:
We should use RFC2119, like the Regs: https://www.worldcubeassociation.org/regulations/


jfly:
Great!

PR #200, opened 2016-01-20
Update SquareOnePuzzle.java

Mollerz:
Square-1 colour scheme change to BOY with Yellow top, Red front.

### DIFF
--- a/scrambles/src/puzzle/SquareOnePuzzle.java
+++ b/scrambles/src/puzzle/SquareOnePuzzle.java
@@ -51,12 +51,12 @@ public class SquareOnePuzzle extends Puzzle {
 
     private static HashMap<String, Color> defaultColorScheme = new HashMap<String, Color>();
     static {
-        defaultColorScheme.put("L", new Color(0xffff00));
-        defaultColorScheme.put("B", new Color(0xff0000));
-        defaultColorScheme.put("R", new Color(0x0000ff));
-        defaultColorScheme.put("F", new Color(0xffc800));
-        defaultColorScheme.put("U", new Color(0xffffff));
-        defaultColorScheme.put("D", new Color(0x00ff00));
+        defaultColorScheme.put("L", new Color(0x0000ff));
+        defaultColorScheme.put("B", new Color(0xffc800));
+        defaultColorScheme.put("R", new Color(0x00ff00));
+        defaultColorScheme.put("F", new Color(0xff0000));
+        defaultColorScheme.put("U", new Color(0xffff00));
+        defaultColorScheme.put("D", new Color(0xffffff));
     }
     @Override
     public HashMap<String, Color> getDefaultColorScheme() {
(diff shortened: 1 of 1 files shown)


jfly:
2016-01-20_1809x729_kaladin

Mollerz:
Looks good!

jfly:
@notkevin888, could you take a look? Thanks!

lgarron:
Since yellow and white are now going to be on the same sides, could we make the yellow more constrasty?

In particular, it would be nice if a typical black-and-white printer will show the yellow clearly.

Mollerz:
Maybe we should consider this for ALL puzzles that use BOY then? I used the same colours as things like 3x3, 4x4 etc.

jfly:
I agree, if we're going to change this here, we should change it everywhere.

lgarron:
I don't strictly disagree, but it's more important for Square-1. We can push the yellow color farther away from yellow for Square-1 than for other puzzles, since it can only show up on the top/bottom.

jfly:
That's an interesting point, @lgarron. Increasing constrast between yellow and white here sounds reasonable to me. @Mollerz, can you pick something you think looks good?

Mollerz:
Went towards a goldenrod sort of colour.

jfly:
Could you add a comment to the code that explains why this yellow differs from our other BOY yellows?

Mollerz:
Done

jfly:
Screenshot:

2016-01-21_1822x791_kaladin

I'm not wild about this change, sorry @Mollerz. I think our orange now looks more yellow than our yellow.

@lgarron, what do you think?

lgarron:
Hmm, this is hard. Best I could come up with was http://leaverou.github.io/contrast-ratio/#%23f90-on-%23fe0 . Better than before, but the yellow is still really light.

068f0cd6-c035-11e5-9191-008ead218e8b

Mollerz:
Yeah I didn't like my change either, it keeps the contrast but looks awful. At this point I'm not entirely sure what to suggest...

jfly:
I'd vote for the original yellow, or @lgarron's yellow in his latest post. I leave the decision up to you, @Mollerz.

Mollerz:
At this point I think keeping consistency and using the original yellow would be the best option. I'd recommend that we let delegates know of the change and see whether the lack of contrast difference between the white/yellow is a problem, if it is we'll work something out.

jfly:
Ok, please change the code to whatever color you're happiest with, and I'll merge this up and try to spin a release.

lgarron:
> or @lgarron's yellow in his latest post

Note that I also changed orange to compensate.

Mollerz:
Done

lgarron:
I just printed out a Square-1 sheet as a test, on a color printer. I can tell the yellow and white apart without too much trouble if I hold the sheet at reading distance and look at the image, but the distinction completely dies if I hold it farther. Distinguishing a white edge in an otherwise yellow section is not going to happen on its own.

PR #201, opened 2016-01-20
Update MegaminxPuzzle.java

Mollerz:
Change megaminx colour scheme from outdated mefferts scheme to mf8/Dayan/Shengshou/YuHu colour scheme.

### DIFF
--- a/scrambles/src/puzzle/MegaminxPuzzle.java
+++ b/scrambles/src/puzzle/MegaminxPuzzle.java
@@ -196,18 +196,18 @@ public class MegaminxPuzzle extends Puzzle {
     @Override
     public HashMap<String, Color> getDefaultColorScheme() {
         HashMap<String, Color> colors = new HashMap<String, Color>();
-        colors.put("U", new Color(0xFFFFFF));
-        colors.put("BL", new Color(0x882222));
-        colors.put("BR", new Color(0x0000FF));
-        colors.put("R", new Color(0x880088));
-        colors.put("F", new Color(0x008800));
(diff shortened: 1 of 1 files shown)


jfly:
2016-01-20_1783x794_kaladin

jfly:
I haven't tried looking at hex colors, but did you darken our second darkest blue?

Mollerz:
Yes, I took the darkblue from 0000ff to 0000b3. I asked some people on their thoughts on the colour changes and they all said the original blue was a bit too bright.

jfly:
And purple stayed the same?

Can you provide some screenshots justifying the change? To my eyes, our dark blue and our purple are much harder to distinguish than they were previously.

Mollerz:
The purple was also made darker. Both the purple and the blue are quite dark on the scheme, the picture attached is in bright lighting. On the screenshot they do look quite similar, however I can personally quite easily distinguish between the colours with f.lux activated right now. I think we could easily make the purple slightly lighter.

img_20160120_214602

jfly:
I believe we should optimize for highest contrast, rather than real life accuracy.

Mollerz:
I increased the lightness of the purple by 30%, try that.

jfly:
That looks *much* better:

2016-01-20_1766x766_kaladin

Mollerz:
Agreed!

jfly:
Looks like the test passed. I'm happy with this. @lgarron, what do you think?

lgarron:
> @lgarron, what do you think?

Looks nice to me!

jfly:
Ok, merged!

PR #214, opened 2016-06-04
Added TNoodle version to FMC sheet. This fixes #213.

### LINKED ISSUE #213: Include TNoodle version in FMC sheet
We do this everywhere else, so for consistency, this would be nice to have.

jfly:
Before:

image

After:

image


SAuroux:
Hey Jeremy,

I support adding the TNoodle Version to the FMC sheets, but your proposal is probably the worst possible place to put it. :) You are basically making it the first point of the competitor instructions, which will surely confuse unfamiliar competitors.

My preference would be on the bottom of the page, like for other events. If that is too hard to do now, please put it on the bottom of the instructions and do not include it in the item list, but rather centered and in brackets.

jfly:
There's no space at the bottom of the page. I'll move it to the bottom of the list.

jfly:
Moved to the bottom!

lgarron:
Just saw this now. The bottom of the list looks like the wrong place – it should not concern competitors, and can only cause confusion. :-/

lgarron:
You can switch to A4 to get space at the bottom. :-P

jfly:
Yes, @SAuroux expressed the same concerns in https://github.com/cubing/tnoodle/pull/214#issuecomment-223747232. Rather than pixel push, I chose to close this issue, don't worry about the TNoodle version showing up on the FMC sheets.

Issue #204, opened 2016-03-08
Generate FMC image only sheet


jfly:
@jbcm627 did this by hand for US Nationals: it's sometime's expensive to print out color, so instead, print out a bunch of black and white FMC sheets, along with a page with a bunch of copies of the scramble and image that organizers can cut out in advance.

@KitClement just aksed for this.

gyroninja:
You are looking for a way to have a sheet full of exclusively normal images (not black and white)?

jbcm627:
For US2015, we printed the following solution sheets:
[FMC_solution_sheets.pdf](https://github.com/cubing/tnoodle/files/445327/FMC_solution_sheets.pdf)

And the following scramble sheets (with actual scrambles, of course):
[FMC_solution_sheets.pdf](https://github.com/cubing/tnoodle/files/445328/FMC_solution_sheets.pdf)

jfly:
Closing! This was fixed in #224.

PR #224, opened 2016-09-03
Added Exclusive Scramble and Solution Sheets for #204

gyroninja:
Added a sheet containing 8 copies of the scramble + scramble image

Added a sheet containing a generic solution sheet for a round (well group to be specific)

### DIFF
--- a/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
+++ b/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
@@ -185,7 +185,7 @@ class ScrambleRequest {
 
         this.colorScheme = scrambler.parseColorScheme(scheme);
     }
-    
+
 
     public List<String> getAllScrambles() {
         ArrayList<String> allScrambles = new ArrayList<String>(Arrays.asList(scrambles));
@@ -194,7 +194,7 @@ class ScrambleRequest {
         }
         return allScrambles;
     }
-    
+
 
     public static ScrambleRequest[] parseScrambleRequests(LinkedHashMap<String, String> query, String seed) throws UnsupportedEncodingException, InvalidScrambleRequestException {
         ScrambleRequest[] scrambleRequests;
@@ -216,7 +216,6 @@ class ScrambleRequest {
         }
         return scrambleRequests;
     }
-
 
     private static PdfReader createPdf(String globalTitle, Date creationDate, ScrambleRequest scrambleRequest) throws DocumentException, IOException {
         azzert(scrambleRequest.scrambles.length > 0);
@@ -314,225 +313,8 @@ class ScrambleRequest {
 
     private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequest scrambleRequest, String globalTitle) throws DocumentException, IOException {
         if(scrambleRequest.fmc) {
-            Rectangle pageSize = doc.getPageSize();
             for(int i = 0; i < scrambleRequest.scrambles.length; i++) {
-                String scramble = scrambleRequest.scrambles[i];
-                PdfContentByte cb = docWriter.getDirectContent();
-                float LINE_THICKNESS = 0.5f;
-                BaseFont bf = BaseFont.createFont(BaseFont.HELVETICA, BaseFont.CP1252, BaseFont.NOT_EMBEDDED);
-
-                int bottom = 30;
-                int left = 35;
-                int right = (int) (pageSize.getWidth()-left);
-                int top = (int) (pageSize.getHeight()-bottom);
-
-                int height = top - bottom;
-                int width = right - left;
-
-                int solutionBorderTop = bottom + (int) (height*.5);
-                int scrambleBorderTop = solutionBorderTop + 40;
-
-                int competitorInfoBottom = top - (int) (height*.15);
-                int gradeBottom = competitorInfoBottom - 50;
-                int competitorInfoLeft = right - (int) (width*.45);
-
-                int rulesRight = competitorInfoLeft;
-
-                int padding = 5;
-
-                // Outer border
-                cb.setLineWidth(2f);
-                cb.moveTo(left, top);
-                cb.lineTo(left, bottom);
-                cb.lineTo(right, bottom);
-                cb.lineTo(right, top);
-
-                // Solution border
-                cb.moveTo(left, solutionBorderTop);
-                cb.lineTo(right, solutionBorderTop);
-
-                // Rules bottom border
-                cb.moveTo(left, scrambleBorderTop);
-                cb.lineTo(rulesRight, scrambleBorderTop);
-
-                // Rules right border
-                cb.lineTo(rulesRight, gradeBottom);
-
-                // Grade bottom border
-                cb.moveTo(competitorInfoLeft, gradeBottom);
-                cb.lineTo(right, gradeBottom);
-
-                // Competitor info bottom border
-                cb.moveTo(competitorInfoLeft, competitorInfoBottom);
-                cb.lineTo(right, competitorInfoBottom);
-
-                // Competitor info left border
-                cb.moveTo(competitorInfoLeft, gradeBottom);
-                cb.lineTo(competitorInfoLeft, top);
-
-                // Solution lines
-                int availableSolutionWidth = right - left;
-                int availableSolutionHeight = scrambleBorderTop - bottom;
-                int lineWidth = 25;
-                //int linesX = (availableSolutionWidth/lineWidth + 1)/2;
-                int linesX = 10;
-                int linesY = (int) Math.ceil(1.0*WCA_MAX_MOVES_FMC / linesX);
-
-                cb.setLineWidth(LINE_THICKNESS);
-                cb.stroke();
-
-//              int allocatedX = (2*linesX-1)*lineWidth;
-                int excessX = availableSolutionWidth-linesX*lineWidth;
-                int moveCount = 0;
-            solutionLines:
-                for(int y = 0; y < linesY; y++) {
-                    for(int x = 0; x < linesX; x++) {
-                        if(moveCount >= WCA_MAX_MOVES_FMC) {
-                            break solutionLines;
-                        }
-                        int xPos = left + x*lineWidth + (x+1)*excessX/(linesX+1);
-                        int yPos = solutionBorderTop - (y+1)*availableSolutionHeight/(linesY+1);
-                        cb.moveTo(xPos, yPos);
-                        cb.lineTo(xPos+lineWidth, yPos);
-                        moveCount++;
-                    }
-                }
-
-                float UNDERLINE_THICKNESS = 0.2f;
-                cb.setLineWidth(UNDERLINE_THICKNESS);
-                cb.stroke();
-
-                cb.beginText();
-                int availableScrambleSpace = right-left - 2*padding;
-                int scrambleFontSize = 20;
-                String scrambleStr = "Scramble: " + scramble;
-                float scrambleWidth;
-                do {
-                    scrambleFontSize--;
-                    scrambleWidth = bf.getWidthPoint(scrambleStr, scrambleFontSize);
-                } while(scrambleWidth > availableScrambleSpace);
-
-                cb.setFontAndSize(bf, scrambleFontSize);
-                int scrambleY = 3 + solutionBorderTop+(scrambleBorderTop-solutionBorderTop-scrambleFontSize)/2;
-                cb.showTextAligned(PdfContentByte.ALIGN_LEFT, scrambleStr, left+padding, scrambleY, 0);
-                cb.endText();
-
-                int availableScrambleWidth = right-rulesRight;
-                int availableScrambleHeight = gradeBottom-scrambleBorderTop;
-                Dimension dim = scrambleRequest.scrambler.getPreferredSize(availableScrambleWidth-2, availableScrambleHeight-2);
-                PdfTemplate tp = cb.createTemplate(dim.width, dim.height);
-                Graphics2D g2 = new PdfGraphics2D(tp, dim.width, dim.height, new DefaultFontMapper());
-
-                try {
-                    Svg svg = scrambleRequest.scrambler.drawScramble(scramble, scrambleRequest.colorScheme);
-                    drawSvgToGraphics2D(svg, g2, dim);
-                } catch (InvalidScrambleException e) {
-                    l.log(Level.INFO, "", e);
-                } finally {
-                    g2.dispose();
-                }
-
-
-                cb.addImage(Image.getInstance(tp), dim.width, 0, 0, dim.height, rulesRight + (availableScrambleWidth-dim.width)/2, scrambleBorderTop + (availableScrambleHeight-dim.height)/2);
-
-                ColumnText ct = new ColumnText(cb);
-
-                int fontSize = 15;
-                int marginBottom = 10;
-                int offsetTop = 27;
-                boolean showScrambleCount = scrambleRequest.scrambles.length > 1;
-                if(showScrambleCount) {
-                    offsetTop -= fontSize + 2;
-                }
-
-                cb.beginText();
-                cb.setFontAndSize(bf, fontSize);
-                cb.showTextAligned(PdfContentByte.ALIGN_CENTER, globalTitle, competitorInfoLeft+(right-competitorInfoLeft)/2, top-offsetTop, 0);
-                offsetTop += fontSize + 2;
-                cb.endText();
-
-                cb.beginText();
-                cb.setFontAndSize(bf, fontSize);
-                cb.showTextAligned(PdfContentByte.ALIGN_CENTER, scrambleRequest.title, competitorInfoLeft+(right-competitorInfoLeft)/2, top-offsetTop, 0);
-                cb.endText();
-
-                if(showScrambleCount) {
-                    cb.beginText();
-                    offsetTop += fontSize + 2;
-                    cb.setFontAndSize(bf, fontSize);
-                    cb.showTextAligned(PdfContentByte.ALIGN_CENTER, "Scramble " + (i+1) + " of " + scrambleRequest.scrambles.length, competitorInfoLeft+(right-competitorInfoLeft)/2, top-offsetTop, 0);
-                    cb.endText();
-                }
-
-                offsetTop += fontSize + marginBottom;
-
-                cb.beginText();
-                fontSize = 15;
-                cb.setFontAndSize(bf, fontSize);
-                cb.showTextAligned(PdfContentByte.ALIGN_LEFT, "Competitor: __________________", competitorInfoLeft+padding, top-offsetTop, 0);
-                offsetTop += fontSize + marginBottom;
-                cb.endText();
-
-                cb.beginText();
-
-                fontSize = 15;
-                cb.setFontAndSize(bf, fontSize);
-                cb.showTextAligned(PdfContentByte.ALIGN_LEFT, "WCA ID:", competitorInfoLeft+padding, top-offsetTop, 0);
-
-                cb.setFontAndSize(bf, 19);
-                int wcaIdLength = 63;
-                cb.showTextAligned(PdfContentByte.ALIGN_LEFT, "_ _ _ _  _ _ _ _  _ _", competitorInfoLeft+padding+wcaIdLength, top-offsetTop, 0);
-
-                offsetTop += fontSize + (int) (marginBottom*1.8);
-                cb.endText();
-
-
-                cb.beginText();
-                fontSize = 11;
-                cb.setFontAndSize(bf, fontSize);
-                cb.showTextAligned(PdfContentByte.ALIGN_CENTER, "DO NOT FILL IF YOU ARE THE COMPETITOR", competitorInfoLeft + (right-competitorInfoLeft)/2, top-offsetTop, 0);
-                offsetTop += fontSize + marginBottom;
-                cb.endText();
-
-                cb.beginText();
-                fontSize = 11;
-                cb.setFontAndSize(bf, fontSize);
-                cb.showTextAligned(PdfContentByte.ALIGN_CENTER, "Graded by: _______________ Result: ______", competitorInfoLeft + (right-competitorInfoLeft)/2, top-offsetTop, 0);
-                offsetTop += fontSize + marginBottom;
-                cb.endText();
-
-                cb.beginText();
-                cb.setFontAndSize(bf, 25f);
-                int MAGIC_NUMBER = 40; // kill me now
-                cb.showTextAligned(PdfContentByte.ALIGN_CENTER, "Fewest Moves", left+(competitorInfoLeft-left)/2, top-MAGIC_NUMBER, 0);
-                cb.endText();
-
-                com.itextpdf.text.List rules = new com.itextpdf.text.List(com.itextpdf.text.List.UNORDERED);
-                rules.add("Notate your solution by writing one move per bar.");
-                rules.add("To delete moves, clearly erase/blacken them.");
-                rules.add("Face moves F, B, R, L, U, and D are clockwise.");
-                rules.add("Rotations x, y, and z follow R, U, and F.");
-                rules.add("' inverts a move; 2 doubles a move. (e.g.: U', U2)");
-                rules.add("w makes a face move into two layers. (e.g.: Uw)");
-                rules.add("A [lowercase] move is a cube rotation. (e.g.: [u])");
-
-                ct.addElement(rules);
-                int rulesTop = competitorInfoBottom+55;
-                ct.setSimpleColumn(left+padding, scrambleBorderTop, competitorInfoLeft-padding, rulesTop, 0, Element.ALIGN_LEFT);
-                ct.go();
-
-                rules = new com.itextpdf.text.List(com.itextpdf.text.List.UNORDERED);
-                rules.add("You have 1 hour to find a solution.");
-                rules.add("Your solution length will be counted in OBTM.");
-                int maxMoves = WCA_MAX_MOVES_FMC;
-                rules.add("Your solution must be at most " + maxMoves + " moves, including rotations.");
-                rules.add("Your solution must not be directly derived from any part of the scrambling algorithm.");
-                ct.addElement(rules);
-                MAGIC_NUMBER = 150; // kill me now
-                ct.setSimpleColumn(left+padding, scrambleBorderTop, rulesRight-padding, rulesTop-MAGIC_NUMBER, 0, Element.ALIGN_LEFT);
-                ct.go();
-
-                doc.newPage();
+                addFmcSolutionSheet(docWriter, doc, scrambleRequest, globalTitle, i);
             }
         } else {
             Rectangle pageSize = doc.getPageSize();
@@ -555,9 +337,9 @@ class ScrambleRequest {
                 // because the scrambles are so uniformly sized.
                 maxScrambleImageWidth = 190;
             }
-            
+
             Dimension scrambleImageSize = scrambleRequest.scrambler.getPreferredSize(maxScrambleImageWidth, maxScrambleImageHeight);
-            
+
             // First do a dry run just to see if any scrambles require highlighting.
             // Then do the real run, and force highlighting on every scramble
             // if any scramble required it.
@@ -578,7 +360,7 @@ class ScrambleRequest {
                     PdfPTable headerTable = new PdfPTable(1);
                     headerTable.setTotalWidth(new float[] { availableWidth });
                     headerTable.setLockedWidth(true);
-                    
+
                     PdfPCell extraScramblesHeader = new PdfPCell(new Paragraph("Extra scrambles"));
                     extraScramblesHeader.setVerticalAlignment(PdfPCell.ALIGN_MIDDLE);
                     extraScramblesHeader.setPaddingBottom(3);
@@ -586,7 +368,7 @@ class ScrambleRequest {
                     if(!dryRun) {
                         doc.add(headerTable);
                     }
-                    
+
                     scrambleNumberPrefix = "E";
                     TableAndHighlighting extraTableAndHighlighting = createTable(docWriter, doc, sideMargins, scrambleImageSize, scrambleRequest.extraScrambles, scrambleRequest.scrambler, scrambleRequest.colorScheme, scrambleNumberPrefix, forceHighlighting);
                     if(dryRun) {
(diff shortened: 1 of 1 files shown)


webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -315,7 +315,9 @@ private static PdfReader createPdf(String globalTitle, Date creationDate, Scramb
     private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequest scrambleRequest, String globalTitle) throws DocumentException, IOException {
         if(scrambleRequest.fmc) {
             Rectangle pageSize = doc.getPageSize();
+            //addSolutionSheet(docWriter, doc, scrambleRequest, globalTitle);
```
jfly: Vestigial code?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -533,7 +535,10 @@ private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequ
                 ct.go();
 
                 doc.newPage();
+
+                addScrambleSheet(doc, scrambleRequest, globalTitle, scramble, cb, i);
```
jfly: I like that you created methods for these instead of adding them to this already huge function. Could you actually pull out the code that was already here into a method as well? If I understand the naming correctly, there are 3 things:

- Solution sheet with scramble (how about `addFmcSolutionSheetWithScramble()`?)
- Generic solution with without scramble (how about `addFmcSolutionSheetWithoutScramble()`?)
- Sheet with many copies of the same FMC scramble (how about `addFmcScrambleCutoutSheet()`?)

The names are kind of lengthy, but I think they're descriptive.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -533,7 +535,10 @@ private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequ
...
             }
+            addSolutionSheet(docWriter, doc, scrambleRequest, globalTitle);
```
jfly: This is the generic FMC solution sheet that does not have a scramble on it, right? Is it possible that we end up calling this method twice if we generate 2 rounds of FMC?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -603,6 +608,253 @@ private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequ
...
+        cb.endText();
+
+        com.itextpdf.text.List rules = new com.itextpdf.text.List(com.itextpdf.text.List.UNORDERED);
+        rules.add("Notate your solution by writing one move per bar.");
+        rules.add("To delete moves, clearly erase/blacken them.");
+        rules.add("Face moves F, B, R, L, U, and D are clockwise.");
+        rules.add("Rotations x, y, and z follow R, U, and F.");
+        rules.add("' inverts a move; 2 doubles a move. (e.g.: U', U2)");
+        rules.add("w makes a face move into two layers. (e.g.: Uw)");
+        rules.add("A [lowercase] move is a cube rotation. (e.g.: [u])");
```
jfly: FYI, I didn't read this method carefully, but this feels like a copy of the FMC sheet code we already had. Can you refactor this to share the code instead? Rethinking my suggestion from above, perhaps a single method called `addFmcSolutionSheet(boolean includeScramble)`?

jfly: I think it's worth trying to share the code because when/if these rules change, I don't want to have to remember to edit them in two places.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -603,6 +608,253 @@ private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequ
...
+        try {
+            Svg svg = scrambleRequest.scrambler.drawScramble(scramble, scrambleRequest.colorScheme);
+            drawSvgToGraphics2D(svg, g2, dim);
+        } catch (InvalidScrambleException e) {
+            l.log(Level.INFO, "", e);
+        } finally {
+            g2.dispose();
+        }
+
+        for (int y = 0; y < 8; y++) {
```
jfly: Nit: Can you pull the number `8` out into a variable? Perhaps call it `SCRAMBLES_PER_FMC_CUTOUT_SHEET`?

Laura-O:
So this is generating a sheet and adding it to the main file?

I am a bit concerned that Delegates might just print the solution sheet because they think it is the proper sheet and forget to print the scramble.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
gyroninja: Yes. The generic sheet **includes** the round (and group) on it.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
gyroninja: You want this as a constant right? It will make things more complicated if I have to resize things based on that number.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -603,6 +388,331 @@ private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequ
         doc.newPage();
     }
 
+    private static void addSolutionSheet(PdfWriter docWriter, Document doc, ScrambleRequest scrambleRequest, String globalTitle, int index) throws DocumentException, IOException {
+        boolean withScrambles = index != -1;
```
jfly: Why is this pluralized? Shouldn't it be `withScramble`?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -603,6 +388,331 @@ private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequ
...
+
+        cb.beginText();
+        cb.setFontAndSize(bf, fontSize);
+        cb.showTextAligned(PdfContentByte.ALIGN_CENTER, globalTitle, competitorInfoLeft+(right-competitorInfoLeft)/2, top-offsetTop, 0);
+        offsetTop += fontSize + 2;
+        cb.endText();
+
+        cb.beginText();
+        cb.setFontAndSize(bf, fontSize);
+        cb.showTextAligned(PdfContentByte.ALIGN_CENTER, scrambleRequest.title, competitorInfoLeft+(right-competitorInfoLeft)/2, top-offsetTop, 0);
```
jfly: This is something like "3x3x3: Fewest Moves Round 1", right? I think that in the interest of a fully generic FMC scramble sheet, we wouldn't say anything like "Round 1" or "Round 2".

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -603,6 +388,331 @@ private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequ
...
+        rules.add("Your solution must not be directly derived from any part of the scrambling algorithm.");
+        ct.addElement(rules);
+        MAGIC_NUMBER = 150; // kill me now
+        ct.setSimpleColumn(left+padding, scrambleBorderTop, rulesRight-padding, rulesTop-MAGIC_NUMBER, 0, Element.ALIGN_LEFT);
+        ct.go();
+
+        doc.newPage();
+    }
+
+    private static void addSolutionSheet(PdfWriter docWriter, Document doc, ScrambleRequest scrambleRequest, String globalTitle) throws DocumentException, IOException {
```
jfly: Can you rename this to something that makes it a bit clearer that no scramble is involved? Something like `addScramblelessSolutionSheet`, `addGenericSolutionSheet`, or `addSolutionSheetWithoutScramble`. Also, it's surprising to me that a generic FMC solution sheet would need a `ScrambleRequest`. I read through the implementation of `addSolutionSheet`, and I think when we're generating a generic sheet, the `ScrambleRequest` is only used to write something like `3x3x3: Fewest Moves Round 1`, which I don't think we should even be putting (because we want this to be a generic sheet!)

jfly: Also, I'd probably add `Fmc` to the method name here just to make it clear that it has to do with FMC, but that's your call.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -603,6 +388,331 @@ private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequ
...
+        addSolutionSheet(docWriter, doc, scrambleRequest, globalTitle, -1);
+    }
+
+    private static void addScrambleSheet(PdfWriter docWriter, Document doc, ScrambleRequest scrambleRequest, String globalTitle, int index) throws DocumentException, IOException {
```
jfly: I feel more strongly about adding `Fmc` to the method name here because TNoodle generates lots of scramble sheets, and this is a pretty specific kind of scramble sheet.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
jfly: While it would be cool to resize everything based on this number, it's okay if stuff isn't computed in terms of this number, I just want to name the variable in order to make it more obvious what this loop is doing.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -603,6 +391,340 @@ private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequ
         doc.newPage();
     }
 
+    private static void addFMCSolutionSheet(PdfWriter docWriter, Document doc, ScrambleRequest scrambleRequest, String globalTitle, int index) throws DocumentException, IOException {
```
jfly: Legitimate question because I don't remember what TNoodle's convention is: should this be "addFmcSolutionSheet"? (lowercase mc in fmc)

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -314,225 +318,9 @@ private static PdfReader createPdf(String globalTitle, Date creationDate, Scramb
...
-                rules.add("Your solution must be at most " + maxMoves + " moves, including rotations.");
-                rules.add("Your solution must not be directly derived from any part of the scrambling algorithm.");
-                ct.addElement(rules);
-                MAGIC_NUMBER = 150; // kill me now
-                ct.setSimpleColumn(left+padding, scrambleBorderTop, rulesRight-padding, rulesTop-MAGIC_NUMBER, 0, Element.ALIGN_LEFT);
-                ct.go();
-
-                doc.newPage();
+                addFMCSolutionSheet(docWriter, doc, scrambleRequest, globalTitle, i);
+                addFMCScrambleSheet(docWriter, doc, scrambleRequest, globalTitle, i);
```
jfly: 1. Do we want to include the generic fmc solution sheet and the separate fmc scramble sheets in the big pdf with every scramble?
2. If no to 1), then we need to generate a separate pdf file for a generic fmc solution sheet and separate pdf files for fmc scramble sheets (plural!).

I believe the answer to 1) should be "no". I hope that this will address @Laura-O's concerns [here](https://github.com/cubing/tnoodle/pull/224#issuecomment-244532072).

From reading this code, it looks like the fmc scramble sheets will end up in our big pdf. Could you instead move this out to `createPdf()` so we only generate separate files for them?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -234,8 +234,12 @@ private static PdfReader createPdf(String globalTitle, Date creationDate, Scramb
         }
 
         doc.open();
-        // Note that we ignore scrambleRequest.copies here.
-        addScrambles(docWriter, doc, scrambleRequest, globalTitle);
+        if (pdfFileName.equals("3x3x3 Fewest Moves Solution Sheet")) {
```
jfly: I'm not very comfortable with this string comparison. I think the problem here is that there's a lot of boilerplate code involved in createing a PDF and it watermarking/closing it. Could we pull those out into 2 separate methods and then push the call to `addGenericFMCSolutionSheet()` up to `requestsToZip()`? If what I'm saying doesn't make sense, feel free to reach out to me on gchat and we can talk about this.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
gyroninja: It does not end up in the big pdf currently.

Laura-O: Thanks for the answer!

@cubing/wrc-team Are we fine with that?

jfly: > It does not end up in the big pdf currently.

@gyroninja, what are you referring to when you say "it"? I just tested this code out, and I can confirm that the separate fmc scramble sheets *do* show up in the big PDF*. You are correct that the generic fmc solution sheet does not show up in the big PDF.

Call stack:

- [`requestsToPdf`](https://github.com/gyroninja/tnoodle/blob/c313716f3cb5eeef1daf55c68df8e55d59bfe556/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L1229)
- [`createPdf`](https://github.com/gyroninja/tnoodle/blob/c313716f3cb5eeef1daf55c68df8e55d59bfe556/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L221)
-  [`addScrambles`](https://github.com/gyroninja/tnoodle/blob/c313716f3cb5eeef1daf55c68df8e55d59bfe556/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L323)

viroulep: Any chance to have a screenshot of both sheets?

I guess usually FMC solution sheets are printed separately, so having them in the big pdf can actually be annoying as you'd want to print everything but them.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -975,6 +1092,42 @@ public static ByteArrayOutputStream requestsToZip(ServletContext context, String
 
         ZipOutputStream zipOut = new ZipOutputStream(baosZip);
         HashMap<String, Boolean> seenTitles = new HashMap<String, Boolean>();
+
+        for(ScrambleRequest scrambleRequest : scrambleRequests) {
+            if (scrambleRequest.fmc) {
```
jfly: nit: I would probably create a boolean called `isFmcBeingHeld` and update it here and then immediately break.

Afterwards, I'd create an `if (isFmcBeingHeld)` block and put all the logic for creating the generic FMC solution sheet there. This has a benefit of decreasing indentation a bit, and making it very clear that we aren't going to create multiple copies of the same `pdf/3x3x3 Fewest Moves Solution Sheet.pdf` file. With the current code, I had to search a bit to find the `break;` at the end of this block.

Just a comment on how I would do it, feel free to ignore!

gyroninja:
@jfly Sorry for taking so long to get back to this. For the unified pdf currently I add the current fmc sheet and a sheet which contains only the scrambles for it (and no portion to fill out the solution). There is also a fmc solution sheet. The fmc solution sheet does not contain the current scramble on it and it is not the unified pdf. Do you want it to be included in the unified pdf (I assume once)?

@viroulep 

Current layout of zip
https://i.gyazo.com/11dd7ad2ab5b0135c0b84c90c716ccc8.png

Current Pages of Fewest Moves Round 1 (which gets included in the unified pdf)
https://i.gyazo.com/255f1424201acccb9a21c0a000962ac5.png
https://i.gyazo.com/1184167c9fdb56feed5bbf2adfdbfed6.png

Current Page of Fewest Move Solution Sheet (which is NOT included in the unified pdf)
https://i.gyazo.com/9e95993f5c9391446d594e91b78fb6c0.png

jfly:
Urg, I think I manged to have a critical typo in my last comment. This was also long enough ago that I could be misremembering my logic.

Here's what I said:

> I believe that these changes should change the contents of our unified pdf with everything in it. I just checked, and currently with your changes, the unified PDF includes the "fmc scramble sheet", which previously did not exist.

Here's what I **meant** to say:

> I believe that these changes should **not** change the contents of our unified pdf with everything in it. I just checked, and currently with your changes, the unified PDF includes the "fmc scramble sheet", which previously did not exist.

(I think that's why I wrote that second sentence comparing the new unified PDF with the old unified PDF.)

I've edited my original post to fix this enormous typo.

jfly:
Upon further reflection, I don't really care what we end up putting in the unified pdf. What's the simplest for the code?

jfly:
Could you include screenshots demonstrating what this all looks like when you run 1 round of FMC with 3 scrambles?

Laura-O:
I think the design of the small scramble sheets should be improved:

1. If I counted correctly, the scramble has 26 moves, and the last move is nearly touching the image. What happens when the scramble is longer? Is there a line break?
2. Furthermore, it looks like there is no margin on the right side. Is this margin large enough? (see [here](http://stackoverflow.com/questions/3503615/what-are-the-minimum-margins-most-printers-can-handle) for an discussion about this topic)
3. The small scramble sheet is supposed to be cut into smaller pieces. Why not add a cutting line to make that easier?

Regarding the blank FMC Solution sheet: I think this sheet should include the round and attempt number. Trusting the competitors to fill these fields out correctly is not a good idea: many even fail to write their WCA ID correctly, so I am sure a large number of these fields will either be blank or filled out incorrectly (e.g. confusing round and attempt).

gyroninja:
* Increased the right margin (by decreasing the scramble image's size)
* Added a dotted line to aid in cutting
Scramble Page
https://i.gyazo.com/b023b79530a0a1d77238ac43335d12b8.png
Real Life Printout (Thunderclap v2 for reference)
https://i.gyazo.com/60e82938c5f67bf6ad829137ff4ff486.png

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -651,6 +651,7 @@ private static void addGenericFmcSolutionSheet(PdfWriter docWriter, Document doc
     private static void addFmcScrambleSheet(PdfWriter docWriter, Document doc, ScrambleRequest scrambleRequest, String globalTitle, int index) throws DocumentException, IOException {
         Rectangle pageSize = doc.getPageSize();
         String scramble = scrambleRequest.scrambles[index];
+        //scramble = "U' U' U' D' U' F' U' U' U' U' U' L' U' U' U' U' U' R' U' U' U' U' U' U' U' U' U'";
```
lgarron: Remove?

gyroninja:
Added missing dotted line for cutting
https://i.gyazo.com/2d31ebbd5f254ac94a3ceedf22e010b2.png

Removed commented out debug for testing the font scaling to make sure everything fit.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -975,6 +1101,45 @@ public static ByteArrayOutputStream requestsToZip(ServletContext context, String
...
         HashMap<String, Boolean> seenTitles = new HashMap<String, Boolean>();
+
+        boolean fmcBeingHeld = false;
+        for(ScrambleRequest scrambleRequest : scrambleRequests) {
+            if (scrambleRequest.fmc) {
+                fmcBeingHeld = true;
+                break;
+            }
+        }
+        if (fmcBeingHeld) {
```
jfly: spacing nit

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -975,6 +1101,45 @@ public static ByteArrayOutputStream requestsToZip(ServletContext context, String
 
         ZipOutputStream zipOut = new ZipOutputStream(baosZip);
         HashMap<String, Boolean> seenTitles = new HashMap<String, Boolean>();
+
+        boolean fmcBeingHeld = false;
+        for(ScrambleRequest scrambleRequest : scrambleRequests) {
+            if (scrambleRequest.fmc) {
```
jfly: spacing nit

jfly:
> Regarding the blank FMC Solution sheet: I think this sheet should include the round and attempt number. Trusting the competitors to fill these fields out correctly is not a good idea: many even fail to write their WCA ID correctly, so I am sure a large number of these fields will either be blank or filled out incorrectly (e.g. confusing round and attempt).
View changes

@Laura-O, you raise a very good point. However, if we're going to bother putting the round and attempt number on these sheets, you might as well just print out the sheets we already have, right? That way, you get round number, attempt number, *and* a copy of the scramble. If organizers want to deal with competitors possibly not filling in the round and attempt numbers correctly, then they are welcome to try it out.

I can imagine that one argument for including the round+attempt number but not the scramble would be to be able to more casually pass out the solution sheets before an attempt (no need to worry about putting it face down so competitors can't see it). We could add that if it's really something you could see yourself using.

jfly:
@gyroninja, this looks good to me. Great work!

I'm going to ping @Laura-O to get her feedback about my last comment before merging this up.

jfly:
Just talked to @Laura-O, she's okay with these changes.

I plan to merge this up on Monday if no one comments over the weekend.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -314,225 +313,9 @@ private static PdfReader createPdf(String globalTitle, Date creationDate, Scramb
...
-                rules.add("Your solution must be at most " + maxMoves + " moves, including rotations.");
-                rules.add("Your solution must not be directly derived from any part of the scrambling algorithm.");
-                ct.addElement(rules);
-                MAGIC_NUMBER = 150; // kill me now
-                ct.setSimpleColumn(left+padding, scrambleBorderTop, rulesRight-padding, rulesTop-MAGIC_NUMBER, 0, Element.ALIGN_LEFT);
-                ct.go();
-
-                doc.newPage();
+                addFmcSolutionSheet(docWriter, doc, scrambleRequest, globalTitle, i);
+                addFmcScrambleSheet(docWriter, doc, scrambleRequest, globalTitle, i);
```
jfly: Actually, on further thought, could we rename this a little further? I find that "fmc scramble sheet" is a bit ambiguous when talking to people. How about "fmc scramble cutout sheet"? I think the word "cutout" adds a lot of value here.

gyroninja:
* Combined sheet only contains the original style fmc sheet
* Two pdfs are generate for each round. The regular fmc sheet and a scramble cut out sheet
* A pdf is generated for a general solution sheet

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -975,6 +1098,83 @@ public static ByteArrayOutputStream requestsToZip(ServletContext context, String
...
+                docWriter.setBoxSize("art", new Rectangle(36, 54, pageSize.getWidth()-36, pageSize.getHeight()-54));
+
+                doc.addCreationDate();
+                doc.addProducer();
+                if(globalTitle != null) {
+                    doc.addTitle(globalTitle);
+                }
+
+                doc.open();
+                addFmcScrambleCutoutSheet(docWriter, doc, scrambleRequest, globalTitle, 0);
```
jfly: Doesn't this 0 look lonely and random to you =)? This only adds the very first scramble of the round (see demo above).

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -975,6 +1098,83 @@ public static ByteArrayOutputStream requestsToZip(ServletContext context, String
...
+                fmcBeingHeld = true;
+
+                String safeTitle = toFileSafeString(scrambleRequest.title);
+                int salt = 0;
+                String tempNewSafeTitle = safeTitle;
+                while(seenTitles.get(tempNewSafeTitle) != null) {
+                    tempNewSafeTitle = safeTitle + " (" + (++salt) + ")";
+                }
+                safeTitle = tempNewSafeTitle;
+                seenTitles.put(safeTitle, true);
```
jfly: This feels weird to me, because this isn't actually the title of the file you put in the zip file (later on, you append " Scramble Cutout Sheet.pdf"). Won't this cause problems later on when generating the non-cutout sheet?

edit: Confirmed by generated a zip at this url: http://localhost:2014/scramble/#competitionName=Scrambles%20for%202017-04-25&rounds=i('eventID'-'333fm'_'round'-'1'_'groupCount'-2_'scrambleCount'-3_'extraScrambleCount'-0_'copies'-1)_('eventID'-'333fm'_'round'-'2'_'groupCount'-2_'scrambleCount'-3_'extraScrambleCount'-0_'copies'-1)!&version=1.0:

```
~/Downloads @kaladin> unzip -l Scrambles\ for\ 2017-04-25.zip 
Archive:  Scrambles for 2017-04-25.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
     2771  2017-04-25 10:53   pdf/3x3x3 Fewest Moves Round 1 Group A Scramble Cutout Sheet.pdf
     2206  2017-04-25 10:53   pdf/3x3x3 Fewest Moves Solution Sheet.pdf
     9746  2017-04-25 10:53   pdf/3x3x3 Fewest Moves Round 1 Group A (1).pdf
      211  2017-04-25 10:53   txt/3x3x3 Fewest Moves Round 1 Group A (1).txt
     9734  2017-04-25 10:53   pdf/3x3x3 Fewest Moves Round 1 Group B.pdf
      214  2017-04-25 10:53   txt/3x3x3 Fewest Moves Round 1 Group B.txt
     9740  2017-04-25 10:53   pdf/3x3x3 Fewest Moves Round 2 Group A.pdf
      210  2017-04-25 10:54   txt/3x3x3 Fewest Moves Round 2 Group A.txt
     9725  2017-04-25 10:54   pdf/3x3x3 Fewest Moves Round 2 Group B.pdf
      216  2017-04-25 10:54   txt/3x3x3 Fewest Moves Round 2 Group B.txt
     2007  2017-04-25 10:54   Scrambles for 2017-04-25.json
     2029  2017-04-25 10:54   Scrambles for 2017-04-25.jsonp
      550  2017-04-25 10:54   Scrambles for 2017-04-25.html
    37943  2017-04-25 10:54   Scrambles for 2017-04-25.pdf
---------                     -------
    87302                     14 files
```

Note how the "... Round 1 Group A (1).pdf" has that unecessary (1) in it.

Also note that Round 1 Group B is missing, as is Round 2 Group A and Round 2 Group B.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -217,8 +217,7 @@ public ScrambleRequest(String title, String scrambleRequestUrl, String seed) thr
         return scrambleRequests;
     }
 
-
-    private static PdfReader createPdf(String globalTitle, Date creationDate, ScrambleRequest scrambleRequest) throws DocumentException, IOException {
+    private static PdfReader createPdf(String pdfFileName, String globalTitle, Date creationDate, ScrambleRequest scrambleRequest) throws DocumentException, IOException {
```
jfly: Why did you add `pdfFileName` as an argument to this function? It looks like it's not used anywhere in this method.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -975,6 +1098,83 @@ public static ByteArrayOutputStream requestsToZip(ServletContext context, String
...
+                doc.close();
+
+                // TODO - is there a better way to convert from a PdfWriter to a PdfReader?
+                PdfReader pdfReader = new PdfReader(pdfOut.toByteArray());
+                byte[] b = new byte[(int) pdfReader.getFileLength()];
+                pdfReader.getSafeFile().readFully(b);
+                zipOut.write(b);
+
+                zipOut.closeEntry();
+                break;
```
jfly: This break is the reason why we're only generating *one* of the many necessary cutout sheets (see demo above).

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -603,6 +385,347 @@ private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequ
...
+        cb.setFontAndSize(bf, fontSize);
+        cb.showTextAligned(PdfContentByte.ALIGN_CENTER, "Graded by: _______________ Result: ______", competitorInfoLeft + (right-competitorInfoLeft)/2, top-offsetTop, 0);
+        offsetTop += fontSize + (marginBottom*(withScramble ? 1 : 5));
+        cb.endText();
+
+        if(!withScramble) {
+            cb.beginText();
+            fontSize = 11;
+            cb.setFontAndSize(bf, fontSize);
+            cb.showTextAligned(PdfContentByte.ALIGN_CENTER, "Please see seperate sheet for scrambles.", competitorInfoLeft + (right - competitorInfoLeft) / 2, top - offsetTop, 0);
```
jfly: seperate -> separate

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -603,6 +385,347 @@ private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequ
...
+
+        cb.setLineDash(3f, 3f);
+        cb.moveTo(left, top - offsetTop + (marginBottom * 3));
+        cb.lineTo(right, top - offsetTop + (marginBottom * 3));
+        cb.stroke();
+
+        final int scramblesPerSheet = 8;
+        for (int y = 0; y < scramblesPerSheet; y++) {
+            cb.beginText();
+            fontSize = 12;
```
jfly: It's not safe to assume that 12 will always fit. We should probably add some newlines and dynamically compute the font size. However, this PR has been dragging on for too long, so let's get this merged up, and we can tackle this in a later PR if necessary. I've created https://github.com/thewca/tnoodle/issues/235 to track this. Can you just add a comment here saying something like

> We've hardcoded this to 12 for now, see https://github.com/thewca/tnoodle/issues/235 for discussion about a better solution.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -603,6 +385,347 @@ private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequ
...
+            if(scrambleRequest.scrambles.length > 1) {
+                cb.showTextAligned(PdfContentByte.ALIGN_LEFT, globalTitle + " - " + scrambleRequest.title + " - Scramble " + (index + 1) + " of " + scrambleRequest.scrambles.length + ":", left, top - offsetTop, 0);
+            } else {
+                cb.showTextAligned(PdfContentByte.ALIGN_LEFT, globalTitle + " - " + scrambleRequest.title + ":", left, top - offsetTop, 0);
+            }
+            offsetTop += fontSize + marginBottom;
+            cb.endText();
+
+            cb.beginText();
+            fontSize = 15;
```
jfly: This looks strange. Why set fontSize to 15, but then call `cb.setFontAndSize` with `scrambleFontSize`?

gyroninja:
@jfly I addressed the three comments you had with the last revision.
Also #235 is a good idea because the regular (old) fmc sheet also has this problem. I attached an image on the relevant issue to show that.

jfly:
LGTM, thanks!

PR #228, opened 2016-11-07
Add copies to each round of scrambles, fixes #227

### LINKED ISSUE #227: Add number of "Copies" for each round
Imagine such a normal case: we have 6 scramblers in 3x3 first round (for 100 people) and 1 scrambler in 3x3 final round (for 12 people).
So 3 copies of scrambles for first round is required and only 1 copy for final round.
Can we add a number field that specifies the copies? Thus we can only print one copy of the whole PDF file instead of printing 3 copies for 3x3 1st round, 1 copy for 3x3 final and so on.
add copies

Baiqiang:
I found the attribute copies of ScrambleRequest and realized it can be implement by simply adding it to the queries.
Here is the screenshot:
add copies

### DIFF
--- a/scrambler-interface/WebContent/scrambler-interface/js/ui.js
+++ b/scrambler-interface/WebContent/scrambler-interface/js/ui.js
@@ -534,6 +534,7 @@ var mark2 = {};
                 var round = tr.getAttribute("data-round");
                 var numSolves = parseInt(tr.getElementsByClassName("num_solves")[0].value, 10);
                 var numExtraSolves = parseInt(tr.getElementsByClassName("num_extra_solves")[0].value, 10);
+                var numCopies = parseInt(tr.getElementsByClassName("num_copies")[0].value, 10);
 
                 var numGroups = parseInt(tr.getElementsByClassName("num_groups")[0].value, 10);
 
@@ -542,7 +543,8 @@ var mark2 = {};
                     round: round,
                     groupCount: numGroups,
                     scrambleCount: numSolves,
-                    extraScrambleCount: numExtraSolves
+                    extraScrambleCount: numExtraSolves,
+                    copies: numCopies
                 };
                 if(includeElement) {
                     roundJson.element = tr;
@@ -594,6 +596,7 @@ var mark2 = {};
                         title: eventName + " " + title,
                         scrambleCount: round.scrambleCount,
                         extraScrambleCount: round.extraScrambleCount,
+                        copies: round.copies,
 
                         event: round.eventID,
                         round: round.round,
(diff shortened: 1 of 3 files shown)


scrambler-interface/WebContent/scrambler-interface/js/ui.js
```diff
@@ -542,7 +543,8 @@ var mark2 = {};
                     round: round,
                     groupCount: numGroups,
                     scrambleCount: numSolves,
-                    extraScrambleCount: numExtraSolves
+                    extraScrambleCount: numExtraSolves,
+                    copies: numCopies
```
jfly: I realize we didn't have this before, but I'm a fan of trailing commas in these situations (so we don't end up with 2 line diff like we did this time). Could you add a trailing comma after `numCopies`?

webscrambles/WebContent/wca/scramblegen.html
```diff
@@ -292,7 +292,7 @@
             var request = {
                 scrambles: scrambles,
                 extraScrambles: extraScrambles,
-                copies: 1, // TODO - add support for this to ui.js
+                copies: sheet.copies,
```
jfly: LOL! We had a comment for this already?? Ahahahaha

jfly:
Awesome! Thanks for taking a look @Baiqiang.

Can you provide an example of what the new urls look like?

scrambler-interface/WebContent/scrambler-interface/js/ui.js
```diff
@@ -971,13 +976,22 @@ var mark2 = {};
                     );
             numExtraSolvesInput.classList.add("num_extra_solves");
 
+            var numCopiesTD = mark2.dom.appendElement(newEventTR, "td");
+            var numCopiesInput = mark2.dom.appendElement(
+                    numCopiesTD,
+                    "input",
+                    { type: "number", value: numCopies, min: 0 }
```
jfly: Do we want to allow setting copies to 0?

jfly:
Also, just out of paranoia, can you confirm that the value of `copies` doesn't affect generating the individual pdfs? Even if copies is set to 0.

Baiqiang:
The url looks like `http://localhost:2014/scramble/#competitionName=Scrambles%20for%202016-11-08&rounds=i('eventID'-'333'_'round'-'1'_'groupCount'-1_'scrambleCount'-5_'extraScrambleCount'-2_'copies'-2)_('eventID'-'333'_'round'-'2'_'groupCount'-1_'scrambleCount'-5_'extraScrambleCount'-2_'copies'-3)_('eventID'-'555bf'_'round'-'1'_'groupCount'-1_'scrambleCount'-3_'extraScrambleCount'-2_'copies'-1)!&version=1.0`

jfly:
LGTM! :shipit: :ship: it.

PR #230, opened 2016-11-19
Duplicate rounds and scrambles Fixes #229

### LINKED ISSUE #229: Duplicate rounds and scrambles
As reported on the delegate list: ([wrc link](https://groups.google.com/d/msg/wca-wrc/6lyiYd-wFW4/d8adiPTHBAAJ) [delegates link](https://groups.google.com/d/msg/wca-delegates/pSILDbac1bs/PsgUPHDhBgAJ)) there is a bug which results in the generation of two identical rounds with identical scramble sequences:

- open TNoodle;
- it automatically brings a round of 3x3, so add another round for 3x3;
- click in the X box (in the right) to remove round 1;
- add another round for 3x3 and there will be 2 rounds named round 2;
- generate the scrambles and open both rounds pdfs to see that they are identical;

campos20:
Simple solution for duplicate #229 and scrambles: instead of removing the round that the user clicked, remove last round. This can be done with the removeLastRound function.

Line 990
changed from
`removeButton.addEventListener("click", removeRound.bind(null, eventID, newEventTR_ID), false);`

to

`removeButton.addEventListener("click", removeLastRound.bind(null, eventID), false);`

Function removeRound deleted, since it's not used anymore.

### DIFF
--- a/scrambler-interface/WebContent/scrambler-interface/css/ui.css
+++ b/scrambler-interface/WebContent/scrambler-interface/css/ui.css
@@ -209,11 +209,6 @@ h1 {
   width: 10em;
 }
 
-#events_rounds_interface #rounds_table td.round_remove button {
-  height: 100%;
-  width: 3em;
-}
-
 #events_rounds_interface #rounds_table td input.num_groups {
   width: 3em;
 }
--- a/scrambler-interface/WebContent/scrambler-interface/js/ui.js
+++ b/scrambler-interface/WebContent/scrambler-interface/js/ui.js
@@ -872,13 +872,6 @@ var mark2 = {};
             return count;
         };
 
-        var removeRound = function(eventID, scrambleID) {
-            roundsTbody.removeChild(document.getElementById(scrambleID));
-            document.getElementById("event_amount_value_" + eventID).value = numCurrentRounds(eventID);
-
-            updateHash();
-        };
-
         var removeLastRound = function(eventID) {
             var rounds = getRounds(true);
             var lastRoundOfEvent = null;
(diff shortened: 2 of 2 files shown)


jfly:
> If the user customizes the round, like editing Groups, Copies, Extra Scrambles, the trick might be discovered, since this will always erase the last one. -@campos20

To avoid that, could we only show the delete button for the final round of each event, rather than every single round?

campos20:
> To avoid that, could we only show the delete button for the final round of each event, rather than every single round?

I think this should do it but, is the removeButton absolutely necessary in that case? The user can remove the last round of a particular Event using the arrow. Perhaps if we ignore the Remove column, the user could control the number of rounds using only the sheet.

jfly:
> Perhaps if we ignore the Remove column, the user could control the number of rounds using only the sheet.

What does it mean to ignore the Remove column?
I also don't understand what "only the sheet" means. What's the sheet?

campos20:
> What does it mean to ignore the Remove column?
> I also don't understand what "only the sheet" means. What's the sheet?

The user can control the number of rounds using the events_table, so there's [no need](https://s22.postimg.org/5hbvirxgh/Issue.png) for the removeButton, or column "Remove" in each round.

To be honest, it seems a bit lazy, but it gets the job done, simplifies the code and the interface. Perhaps this will make the page little less friendly to the user.

jfly:
That sounds like a great solution to me @campos20. Let's remove all code related to those X buttons.

jfly:
This is looking good! Could you [squash your commits](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)?

jfly:
I squashed the commits myself.

PR #231, opened 2016-12-20
Keep TNoodle version in FMC sheets fixes #213

### LINKED ISSUE #213: Include TNoodle version in FMC sheet

campos20:
In FMC sheets, Header is avoided, but Footer is now kept. To avoid that TNoodle version collides with the area where the competitor writes the solution, the position of the footer was changed and now it can be controlled with the variable OFFSET_FOOTER.


jfly:
Looks like the build is failing. Could you include a before and after screenshot of what your changes do?

campos20:
It does affect more than just the FMC sheet, since it's changing the place of TNoodle version for all events.

This is how it looks like, currently:
https://postimg.org/image/rhbtkgc4f/

This is the proposal
https://postimg.org/image/h5zgrskf3/

jfly:
Hmm... I'm not wild about how far down this is on regular scramble sheets:

# new
image

# original
image

It also seems pretty darn close to the line on FMC sheets:

image

Can we at least get the text to not be touching the line?

campos20:
Well, I think you are right in two ways:

1: if the last line is far down, this may cause trouble during printing
2: indeed, the text touching the line is a little annoying

The new variable OFFSET_FOOTER can change the position of the text (40 was the original, although [there was no variable for it](https://github.com/thewca/tnoodle/blob/master/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L287), just the value).

Should we keep the version where it was before and shrink a bit the area where the competitor writes the solution, by making the lines a bit closer? Reducing text? Reducing some fonts?

Also, I'm not sure how helpful this information is:

> ' inverts a move; 2 doubles a move. (e.g.: U', U2)

And from my perspective, one of the following could be omitted

> Rotations x, y, and z follow R, U, and F.
> A [lowercase] move is a cube rotation. (e.g.: [u])

jfly:
> Should we keep the version where it was before and shrink a bit the area where the competitor writes the solution, by making the lines a bit closer? Reducing text? Reducing some fonts?

How about we reduce the space between the 2 paragraphs of instruction, and then move everything up a bit?

image

I could also see shrinking the line height of the solution area a bit, but not more than 1 or 2 pixels each row.

Laura-O:
All of the information about the notation is **very** helpful and we should **not** omit it. We already have enough trouble with competitors misunderstanding the notation or simply not knowing it, so it would be really careless to remove it.

campos20:
@Laura-O thanks for that! You changed my mind about it.

@jfly I will execute what you've written and post the result here when it's done. Meanwhile, this was already done (before I read your last comment).

captura de tela de 2016-12-20 18-12-09

It shows official / proposal

Basically, it keeps the Version where it was before (for all events) and shrinks a bit the area for the competitor (actually, this was done by increasing the [bottom](https://github.com/thewca/tnoodle/blob/master/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L324)), since the vertical size of the FMC area is [calculated](https://github.com/thewca/tnoodle/blob/master/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L375).

campos20:
After running dozens of tests, this is the result:

@jfly suggestion (remove space between paragraphs and move everything up a bit)
fmc1

Official / Changed

This was quite painful to code. The original code is a little obscure in this part (I'm saying that with love, like Michael Jackson said in "This is It" :smile:).

After running more tests, I don't think that remove space is really necessary. Using the suggestion to move everything up, looks like this.

fmc2

Official / Changed again

So, how do you think we should proceed?

jfly:
The last one looks good to me! First, please fix travis. After that, I'll email the board and delegates asking for their feedback on your latest screenshot.

jfly:
Cool! Tests pass, can you squash your 3 commits now?

PR #236, opened 2017-09-08
Add support for consuming WCIF

jfly:
**Please download a jar file to test this out [here](https://www.dropbox.com/s/yah5ncfjmvoxuvu/TNoodle-WCA-0.12.0-77-gdaa9b61b.jar?dl=1)!**

A gif is worth a thousand pictures:

2017-12-21_13 25 02_breq

Now that https://github.com/thewca/worldcubeassociation.org/pull/1856 has gone live, TNoodle can finally be able to figure out exactly what scrambles you need for a competition, rather than asking for them.

This PR adds a new, React Redux based UI to TNoodle. You can log in via WCA OAuth and be presented with a list of a upcoming competitions that you have the power to generate scrambles for. (If you're offline or just don't want to use the new UI, you can still access the old UI.)

There's still a bit of work left before this is ready to be merged up, I'll be tracking the remaining work here:

- [x] Teach people about the `staging=true` url parameter for testing stuff out

After we merge up this PR, we'll tackle this other issues (just for polish purposes):

- #250 
- #251

PR #241, opened 2017-10-31
Added a glaring warning about clicking the scramble button twice.

jfly:
Some people don't realize that TNoodle generates scrambles in memory, and that confusingly enough, clicking "Scramble!" multiple times doesn't actually generate new scrambles. I've implemented an obnoxious message to teach people how this works. I'm definitely open to suggestions for improving this.

## Before clicking "Scramble!"

2017-10-31_13 00 28_1201x928_breq

## After clicking "Scramble!"

2017-10-31_13 00 35_1194x927_breq

### DIFF
--- a/scrambler-interface/WebContent/scrambler-interface/css/ui.css
+++ b/scrambler-interface/WebContent/scrambler-interface/css/ui.css
@@ -92,6 +92,10 @@ h1 {
   margin-top: 15px;
 }
 
+.scrambleButton.clicked {
+  background-color: red;
+}
+
 #top_interface #helpLinkDiv {
   position: absolute;
   right: 30px;
--- a/scrambler-interface/WebContent/scrambler-interface/js/ui.js
+++ b/scrambler-interface/WebContent/scrambler-interface/js/ui.js
@@ -689,7 +689,7 @@ var mark2 = {};
                     (generatedScrambleCountByPuzzle[sheet.puzzle] || 0) +
                     additionalScramblesCount
                 );
-                        
+
             }
             return generatedScrambleCountByPuzzle;
         }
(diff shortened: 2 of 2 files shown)


lgarron:
Well, that's ugly enough to make anyone notice. :-P

As a short-term workaround, what about hiding the "Scramble" button?

A better approach might be if the scramble button adds rows to a "generated scramble sets" table, and those rows have the actual download links.

jfly:
> As a short-term workaround, what about hiding the "Scramble" button?

That could work, although I think it would confuse people more than adding the note as I've done here.

> A better approach might be if the scramble button adds rows to a "generated scramble sets" table, and those rows have the actual download links.

Hmm... I don't see how that would solve the problem. We're trying to teach people that each of these generated scramble sets share a lot of the same scrambles.

lgarron:
> That could work, although I think it would confuse people more than adding the note as I've done here.

How so? You can still show a message, but it removing the button should certainly prevent the possible confusion from a button that can be pressed twice.

> Hmm... I don't see how that would solve the problem. We're trying to teach people that each of these generated scramble sets share a lot of the same scrambles.

If they share the same scrambles, we should not not add a new row for them. ;-)

How easy is it to make the button do the safest possible thing and regenerate all scrambles every time you press it?

jfly:
> How so? You can still show a message, but it removing the button should certainly prevent the possible confusion from a button that can be pressed twice.

Oh, I wasn't aware you were planning to show a message as well. Sure, that could be a fine solution, but is a little annoying if someone generates scrambles, clicks the button, and then realizes they made a mistake (maybe they didn't generate enough groups or they didn't fill in the competition name). They'll be forced completely regenerate scrambles.

> If they share the same scrambles, we should not not add a new row for them. ;-)

Hmm, I don't think I'm understanding how this would work, then. Could you explain what would happen when you click the scramble button and what would happen if you add a round after that and click the scramble button again?

lgarron:
> They'll be forced completely regenerate scrambles.
> 
>> If they share the same scrambles, we should not not add a new row for them. ;-)
> 
> Hmm, I don't think I'm understanding how this would work, then. Could you explain what would happen when you click the scramble button and what would happen if you add a round after that and click the scramble button again?

They'd have to refresh. ;-)

jfly:
I'm going to close this PR without merging it. We can revisit this after the new UI introduced in #236 goes live.

Issue #177, opened 2014-10-02
Handle non-ascii characters


jfly:
Sébastien just tried to generate scrmables for "Sébra Open", and the pdf title was "SÃ©bra Open".

campos20:
I think that if #159 is implemented, it may fix this as well.

campos20:
When encoding for Internationalization files were in ISO-8859-1, everything worked fine. Setting to UTF-8, this very same problem happened in FMC translation.

ISO
![captura de tela de 2017-11-14 20-27-49](https://user-images.githubusercontent.com/10170850/32808512-645d085e-c97a-11e7-86e4-b8ade8139c28.png)

UTF8
![captura de tela de 2017-11-14 20-27-55](https://user-images.githubusercontent.com/10170850/32808517-6aba6de0-c97a-11e7-90e9-612dad35d4d2.png)

jfly:
@campos20, yikes! Let's continue this discussion over on #242. This issue is for non-ascii characters in the filename, whereas you're running into trouble with non-ascii characters in the pdfs.

campos20:
I try this from time to time with some new idea, but it's a lot of chained stuff to dig in, but I think I found something perhaps useful. When I open http://localhost:2014 (let's say legacy) and save the .html to my computer, this is what I get

```
<!DOCTYPE html>
<!-- saved from url=(0188)http://localhost:2014/scramble-legacy/#competitionName=T%C3%A9st&rounds=i('eventID'-'333'_'round'-'1'_'scrambleSetCount'-1_'scrambleCount'-5_'extraScrambleCount'-2_'copies'-1)!&version=1.0 -->
<html slick-uniqueid="3"><head><meta http-equiv="Content-Type" content="text/html; **charset=windows-1252**">
<title>WCA Scramble Program</title>
```

I tried to find where this was generated, but I got nothing yet. I believe charset should be set to utf-8.

campos20:
This [code](https://github.com/campos20/tnoodle/blob/3e0a07e82eed7d817b3e156e7dc9588f8d3d27ca/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleViewHandler.java#L247-L248) "fixed" encoding for PDFs, using legacy ui

suushiemaniac:
This is fixed for the PDF library and file encodings now, but interestingly the ZIP filenames can be broken under certain circumstances. Keeping this open to continue investigation

suushiemaniac:
[...]

PR #242, opened 2017-11-10
Internationalization support in FMC

### LINKED ISSUE #109: Translate the rules in fmc scramble
https://github.com/cubing/tnoodle/blob/master/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L496

These rules(or notice) are written in English, should we translate them into different languages?

campos20:
Community can easily help translating FMC to some specific language. All someone needs to do is to include a file in webscrambles/Internationalization/ translating the lines from MessagesBundle.properties, saving the files as MessagesBundle_laguageX_countryY.properties

(this fixes #109 --jfly)


campos20:
I think Travis needs to be update to avoid checking the internationalization files. Travis doesn't like utf8.

jfly:
Are you sure the files are in utf8? The error from Travis makes it sound
like the files are not actually encoded in utf8.

lgarron:
Sounds like pressure to figure out https://github.com/thewca/wca-regulations/issues/456 :-D

campos20:
As far as I can tell, bundles file with Portuguese translation, which contains some É, Ã, Ó, is triggering the failure.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -1233,6 +1282,118 @@ public static ByteArrayOutputStream requestsToZip(ServletContext context, String
...
+
+                doc.addCreationDate();
+                doc.addProducer();
+                if(globalTitle != null) {
+                    doc.addTitle(globalTitle);
+                }
+
+                doc.open();
+                for (int i=0; i<scrambleRequest.scrambles.length; i++){
+                    addFmcSolutionSheet(docWriter, doc, scrambleRequest, globalTitle, i, language, country);
```
jfly: It looks painful to pass around language *and* country everywhere you go. How about just passing around a Locale object instead?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -613,69 +632,99 @@ private static void addFmcSolutionSheet(PdfWriter docWriter, Document doc, Scram
...
+            // a better approach would be considering the actual rendered font height
+            if (totalHeight < (0.65*rulesRectangle.getHeight())){
+                break;
+            }
+            potentialFontSize -= 0.1f;
+        }
+        
+        String rules = "";
+        for (String item : rulesList){
+            rules += item+"\n";
```
jfly: This feels like you could do a string join here instead of a loop

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -613,69 +632,99 @@ private static void addFmcSolutionSheet(PdfWriter docWriter, Document doc, Scram
...
-        int maxMoves = WCA_MAX_MOVES_FMC;
-        rules.add("Your solution must be at most " + maxMoves + " moves, including rotations.");
-        rules.add("Your solution must not be directly derived from any part of the scrambling algorithm.");
-        ct.addElement(rules);
-        MAGIC_NUMBER = 150; // kill me now
-        ct.setSimpleColumn(left+padding, scrambleBorderTop, rulesRight-padding, rulesTop-MAGIC_NUMBER, 0, Element.ALIGN_LEFT);
+        Rectangle rulesRectangle = new Rectangle(left+padding, scrambleBorderTop, competitorInfoLeft-padding, rulesTop);
+        
+        float potentialFontSize = 15f;
+        // resize rules vertically
```
jfly: Can this logic go into fitText instead?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -557,49 +566,59 @@ private static void addFmcSolutionSheet(PdfWriter docWriter, Document doc, Scram
         if(showScrambleCount) {
             offsetTop -= fontSize + 2;
         }
+        
+        String tempMessage = ""; // saves typing in i18n
```
jfly: I don't understand the purpose of this variable or the comment here, could you explain what it's about?

campos20: FitText requires the whole string to compute size, then we have to render that very same string. Instead of typing it [again](https://github.com/campos20/tnoodle/blob/master/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L576) and [again](https://github.com/campos20/tnoodle/blob/master/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L577), tempMessage holds the string and can be used twice without verbose typing.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: That was my first try. I came into some trouble with line breaks and the part of the function that assumes [we are fitting a scramble](https://github.com/campos20/tnoodle/blob/master/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L828). I decided to emulate this function using some ideas from there.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: Ok :)

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: Allright

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: I'll fix this.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -1233,6 +1282,118 @@ public static ByteArrayOutputStream requestsToZip(ServletContext context, String
             zipOut.putNextEntry(null, parameters);
             zipOut.write(join(stripNewlines(scrambleRequest.getAllScrambles()), "\r\n").getBytes());
             zipOut.closeEntry();
+                
+            parameters.setFileNameInZip("Internationalization/");
```
jfly: What's the point of this line?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -1233,6 +1282,118 @@ public static ByteArrayOutputStream requestsToZip(ServletContext context, String
...
+                
+                if (temp.length != 4){
+                    continue;
+                }
+                
+                String language = temp[1];
+                String country = temp[2];
+                
+                // fewest moves regular sheet
+                pdfFileName = "Internationalization/"+language+"_"+country+"_"+safeTitle+".pdf";
```
jfly: These are PDFs, so I believe they belong in the pdf folder. I also think it's weird to capitalize the folder name when non of our other folders are capitalized. How about something like `"pdfs/i18n/"+language+"_"+country+"_"+safeTitle+".pdf"` instead?

jfly:
(Continuing the discussion from https://github.com/thewca/tnoodle/issues/177#issuecomment-344421191)

As for the unicode encoding issue, it turns out that java property files are interpreted in ISO-8859-1 by default. See https://stackoverflow.com/a/4660195 for a good explanation. I was able to get things working by adding the `UTF8Control` class as they suggested and changing the code that calls `ResourceBundle.getBundle` to use this `UTF8Control` class.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: I agree with you about the capitalization and folder placing. I don't think i18n is a good name, since is very common among programmers, but not to non programmers. I'm afraid the UI might get less user friendly.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: I set it to create the folder when I was learning about how tnoodle handled pdfs. I'll supress the line.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: I'm considering pdf/translations. Is intuitive and short enough.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -418,7 +420,14 @@ private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequ
...
     }
 
-    private static void addFmcSolutionSheet(PdfWriter docWriter, Document doc, ScrambleRequest scrambleRequest, String globalTitle, int index) throws DocumentException, IOException {
+    private static void addFmcSolutionSheet(PdfWriter docWriter, Document doc, ScrambleRequest scrambleRequest, String globalTitle, int index, Locale locale) throws DocumentException, IOException {
+        
+        // internationalization
+        Locale currentLocale;
+        ResourceBundle messages;
+        currentLocale = new Locale(locale.getLanguage(), locale.getCountry());
+        messages = ResourceBundle.getBundle("net.gnehzr.tnoodle.server.webscrambles.Internationalization.MessagesBundle", currentLocale);
```
jfly: Could you pull this logic out into a `translate(key, locale)` method that we then use below (instead of the `messages.getString` stuff? (Also, it seems silly to declare the variables and then initialize them on the next lines.)

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
jfly: Ahh, I see now! Could we create a method to handle this for us? Maybe a `fitAndShowText()` method?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -801,90 +833,93 @@ public static float fitText(Font font, String text, Rectangle rect, float maxFon
...
-            if(endIndex < paddedScramble.length()) {
-                while(true) {
-                    if(endIndex < startIndex) {
-                        // We walked all the way to the beginning of the line
-                        // without finding a good breaking point. Give up and break
-                        // in the middle of a turn =(.
-                        endIndex = perfectFitEndIndex;
+        String[] lineList = text.split("\n");
+        
+        for (String line:lineList){
```
jfly: spacing nit: `){` -> `) {`

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -801,90 +833,93 @@ public static float fitText(Font font, String text, Rectangle rect, float maxFon
...
-            startIndex = endIndex;
-            Chunk lineChunk = new Chunk(scrambleSubstring);
-            lineChunks.add(lineChunk);
-            lineChunk.setFont(scrambleFont);
-
-            // Force a line wrap!
-            lineChunk.append("\n");
         }
+        
+//        for (Chunk c : lineChunks) System.out.println(c); // print chunks to check
```
jfly: Vestigial code?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -1195,7 +1230,7 @@ public static ByteArrayOutputStream requestsToZip(ServletContext context, String
             }
 
             doc.open();
-            addGenericFmcSolutionSheet(docWriter, doc, globalTitle);
+            addGenericFmcSolutionSheet(docWriter, doc, globalTitle, new Locale("en", "US"));
```
jfly: Why don't we generate a generic fmc solution sheet for every language?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -769,22 +798,25 @@ private static void addFmcScrambleCutoutSheet(PdfWriter docWriter, Document doc,
...
                 // one line, then potentialFontSize is too large.
                 maxFontSize = potentialFontSize;
             } else {
                 // The font size seems to be a pretty good estimate for how
                 // much vertical space a row actually takes up.
-                float totalHeight = lineChunks.size() * potentialFontSize;
+                
+                // 1.5 due line spacing
+                // hack to fit rules in fmc and not mess with anything else (particularly big cubes)
+                float totalHeight = (lineSpacing ? 1.5f : 1f) * lineChunks.size() * potentialFontSize;
```
jfly: As discussed with you over IM, I think we should ask the PDF for it's leading here, rather than hard coding 1.5. We should also check if the input text has newlines, rather than requiring people calling this function to pass in a lineSpacing boolean.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: Done.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: Yes! Removed.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: Well, we do

https://github.com/campos20/tnoodle/blob/master/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L1378

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -1233,6 +1268,123 @@ public static ByteArrayOutputStream requestsToZip(ServletContext context, String
...
+                doc.close();
+
+                pdfReader = new PdfReader(pdfOut.toByteArray());
+                b = new byte[(int) pdfReader.getFileLength()];
+                pdfReader.getSafeFile().readFully(b);
+                zipOut.write(b);
+
+                zipOut.closeEntry();
+                
+                // Cutout sheet. Is this really necessary?
```
jfly: I agree, this doesn't seem to be necessary, since it's not language specific.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: I agree about the silliness of this declaration.

This translate(key, locale) can easily be added, but what's the point?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: Removed

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: I did not find it. Are you looking at the removed line?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/Internationalization/MessagesBundle.properties
```diff
@@ -0,0 +1,23 @@
+Scramble = Scramble
```
jfly: Why does this exist in addition to MessagesBundle_en_US.properties?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
jfly: No, I'm looking at line 834 here:

image

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -801,91 +825,92 @@ public static float fitText(Font font, String text, Rectangle rect, float maxFon
         return potentialFontSize;
     }
 
-    private static LinkedList<Chunk> splitScrambleToLineChunks(String paddedScramble, Font scrambleFont, float scrambleColumnWidth) {
+    private static LinkedList<Chunk> splitTextToLineChunks(String text, Font scrambleFont, float scrambleColumnWidth) {
```
jfly: Since you renamed scramble to text in the method name, shouldn't we also rename all the other occurrences of scramble (`scrambleFont` and `scrambleColumnWidth`)?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -801,91 +825,92 @@ public static float fitText(Font font, String text, Rectangle rect, float maxFon
...
-                        // in the middle of a turn =(.
-                        endIndex = perfectFitEndIndex;
+        String[] lineList = text.split("\n");
+        
+        for (String line:lineList){
+            int startIndex = 0;
+            int endIndex = 0;
+            while(startIndex < line.length()) {
+                // Walk forwards until we've grabbed the maximum number of characters
+                // that fit in a line, we've run out of characters, or we hit a newline.
```
jfly: This comment says "... we hit a newline", but I don't see the code checking anywhere for a newline, it just seems to be checking if we've used up all the characters or if we've used up all the horizontal space.

I see that you added a split above that separates on newlines. Would it be cleaner to add logic here to check for newline instead? I feel like that would reduce the amount of indentation going on in this method. Alternatively, if that doesn't work, can we fix this comment so it's accurate?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -418,7 +420,12 @@ private static void addScrambles(PdfWriter docWriter, Document doc, ScrambleRequ
         doc.newPage();
     }
 
-    private static void addFmcSolutionSheet(PdfWriter docWriter, Document doc, ScrambleRequest scrambleRequest, String globalTitle, int index) throws DocumentException, IOException {
+    private static void addFmcSolutionSheet(PdfWriter docWriter, Document doc, ScrambleRequest scrambleRequest, String globalTitle, int index, Locale locale) throws DocumentException, IOException {
+        
+        // internationalization
+        Locale currentLocale = new Locale(locale.getLanguage(), locale.getCountry());
+        ResourceBundle messages = ResourceBundle.getBundle("net.gnehzr.tnoodle.server.webscrambles.Internationalization.MessagesBundle", currentLocale);
```
jfly: it looks this code still suffers from the unicode encoding issue. I proposed a fix for that here: https://github.com/thewca/tnoodle/pull/242#issuecomment-344471075.

jfly:
FYI, I added a commit directly to your branch to handle searching for available properties files. See https://github.com/thewca/tnoodle/pull/242/commits/1a58c85d2040627e2f949c6340daf08b61507269.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: Spliting into lines makes this function fit for single and multi line text, so we can use it in rules, megaminx scrambles and single lines text. I've fixed the comments and kept the logic :smile:

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/Internationalization/MessagesBundle.properties
campos20: Removed. I just assumed this was some standard proceeding in i18n (https://docs.oracle.com/javase/tutorial/i18n/intro/after.html). It's working fine without the file.

campos20:
Megaminx is ok. I tested all the puzzles to check.

captura de tela de 2017-11-26 00-28-19

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -730,15 +745,21 @@ private static void addFmcScrambleCutoutSheet(PdfWriter docWriter, Document doc,
...
-                cb.showTextAligned(PdfContentByte.ALIGN_LEFT, globalTitle + " - " + scrambleRequest.title + " - Scramble " + (index + 1) + " of " + scrambleRequest.scrambles.length + ":", left, top - offsetTop, 0);
+                title = globalTitle + " - " + scrambleRequest.title + " - Scramble " + (index + 1) + " of " + scrambleRequest.scrambles.length + ":";
             } else {
-                cb.showTextAligned(PdfContentByte.ALIGN_LEFT, globalTitle + " - " + scrambleRequest.title + ":", left, top - offsetTop, 0);
+                title = globalTitle + " - " + scrambleRequest.title + ":";
             }
-            offsetTop += fontSize + marginBottom;
-            cb.endText();
+            
+            fontSize = (int)fitText(new Font(bfBold), title, new Rectangle(availableScrambleSpace, 100), scrambleFontSize, false, 1f);
```
jfly: Why the case to `int` here? Why not change `fontSize` to a `float`?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -730,15 +745,21 @@ private static void addFmcScrambleCutoutSheet(PdfWriter docWriter, Document doc,
...
 
+            cb.setFontAndSize(bfBold, fontSize); // TODO: Dynamic font size: #235
```
jfly: Is the comment here vestigial?

jfly:
I just updated the code to use yml files for translations. This avoids the weird utf8 issue we were experiencing with properties files, and should allow translators to use @jonatanklosko's app (see https://github.com/thewca/worldcubeassociation.org/wiki/Translating-the-website#internationalize).

jfly:
Ok, I think we can merge this up! I've created https://github.com/jonatanklosko/internationalize/pull/7 to track adding support to https://internationalize.herokuapp.com/ for various kinds of yaml indentation.

jfly:
Merged up. Thanks @campos20!!

PR #246, opened 2017-12-05
add chinese translation for fmc (CJK fonts needed)

cs0x7f:
It seems that the arialuni.ttf does not support most of Chinese characters. 
Furthermore, even if we find a font that support Chinese, the same problem might still occur when we add other languages, e.g. Japanese, Korean, and especially some other minority languages. 
Should we use different ttf files for different languages, or find/generate a single ttf file to support all languages if possible?


jfly:
Thanks, @cs0x7f! I think having a single font with everything in it would be simplest, and keep the TNoodle jar file small. However, as you said, that might be difficult to find. Do you have a ttf file that works for this translation you provided?

cs0x7f:
[wqy-microhei.ttf.zip](https://github.com/thewca/tnoodle/files/1533934/wqy-microhei.ttf.zip)
Attached. 
Wen Quan Yi is an open-source font and used by many open-source os, e.g. Ubuntu, Debian, etc

jfly:
Cool! The file you attached seems to work well for Chinese *and* all our other existing translations. Here's a screenshot of the `zh-CH` translation you submitted:

image

I tried to read up on Wen Quan Yi, and [wikipedia says](https://en.wikipedia.org/wiki/WenQuanYi#Micro_Hei):

> WenQuanYi Micro Hei (文泉驿微米黑), WenQuanYi Micro Hei Mono (文泉驿等宽微米黑) are derived from the Droid Sans font (merged with Droid Sans Fallback) and readable in compact sizes. The primitive motivation of this project was to extend Droid Sans Fallback's glyph coverage. Since the GB 18030 compatible Droid Sans Fallback font's release, the Micro Hei project has been de facto inactive.

Should we be considering a different Wen Quan Yi font that is not "de facto inactive"?

cs0x7f:
For Chinese characters, it does not matter, since almost all frequently used Chinese characters have already been included. While if we are going to support more languages, e.g. Japanese or Korean, Noto CJK fonts designed by Google (https://www.google.com/get/noto/help/cjk/) might be a better choice.

jfly:
Ok, sounds good! Thanks for the link to CJK by Google, we can definitely look into those if people send us translations for Korean or Japanese =)

PR #254, opened 2017-12-25
Update FMC sheet moves.

lgarron:
Formatting/layout is still uneven, and I *really* want to implement a table layout before shipping.

However, I think this is a reasonable middle ground, and already updates all the strings so that we can get translations.

screen shot 2017-12-24 at 19 09 53

The "Fewest Moves" title is given a little less space, so that the text formatting in English doesn't have a lot of awkward space at the bottom.

### DIFF
--- a/webscrambles/src/i18n/en.yml
+++ b/webscrambles/src/i18n/en.yml
@@ -10,16 +10,16 @@ en:
         graded: Graded by
         result: Result
 
-        rule1: Notate your solution by writing one move per bar.
-        rule2: To delete moves, clearly erase or blacken them.
-        rule3: Face moves F, B, R, L, U, and D are clockwise.
-        rule4: Rotations x, y, and z follow R, U, and F.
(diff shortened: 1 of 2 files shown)


jfly:
This looks great! I've created #255 to track implement an actual table based layout.

@Laura-O, could you take a look at this to confirm that things look good from the WRC perspective? After that, I'll merge this up, and we can get started asking for translations.

campos20:
It looks like we are not allowing wide moves anymore (Rw, Fw...). Although FMers avoid them, beginners  using CFOP or Roux might be confused about how to adapt an algorithm. They are also good replacemente for inner layers (M = Rw' R). Perhaps I'll have less people interested in this event.

Laura-O:
The screenshot looks good!  Thanks @lgarron!

jfly:
Ok, I'm going to merge this up!

@campos20, let's continue this discussion over on https://github.com/thewca/wca-regulations/issues/456, where the idea for this PR originated.

PR #257, opened 2017-12-30
Add Japanese translation!

jfly:
Thanks to WCA Delegate [Yuki Tanaka (田中悠樹)](https://www.worldcubeassociation.org/persons/2010TANA02) for translating this.

image


jfly:
Yuki is pleased with this translation, so I'm merging it up now.

PR #262, opened 2017-12-30
Add Russian!

jfly:
Thanks to WCA Delegate [Oleg Gritsenko](https://www.worldcubeassociation.org/persons/2011GRIT01) for the translation!

image


jfly:
Oleg confirmed that this translation looks good, so I'm merging it up.

PR #267, opened 2018-01-01
Update Polish translation

jonatanklosko:
I talked to a delegate and he said it looks good.
The long translations of *clockwise* and *anti-clockwise* break the layout a bit, but these are the only accurate phrases.

image


jonatanklosko:
Wojciech Szatanowski was the delegate =)

jfly:
@jonatanklosko, as with the WCA website, feel free to merge this up when you're happy with it!

PR #270, opened 2018-01-02
Update fr translation

viroulep:
Preview:
fm

@julesDesjardin, @HippolyteM, @zeecho: any chance you could proofread the new version please?
I've proofread it several times but you never know...

### DIFF
--- a/webscrambles/src/i18n/fr.yml
+++ b/webscrambles/src/i18n/fr.yml
@@ -18,33 +18,36 @@ fr:
         graded: Corrigé par
         #original_hash: 5faa59d
         result: Résultat
-        #original_hash: 4c41d93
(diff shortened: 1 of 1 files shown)


zeecho:
Seems good to me

julesDesjardin:
Looks good to me too !

HippolyteM:
I would have just replaced "tiret" with "emplacement" for clarity sake, but it's not absolutely necessary.
(also "noircissez" may not be the prettiest word. "Rayez" would sounds more natural, even if "noircissez" is the closest translation of "blacken", in this context I find it weird)

But I may be too late.:/

viroulep:
Never too late, see #274

PR #274, opened 2018-01-04
Fix wording in French

viroulep:
Changes suggested by @HippolyteM.

I'm still discussing offline with him about the other suggestions.
This is purely French nitpick, so I'll merge this as soon as we come to a consensus.


viroulep:
We agreed :)

PR #276, opened 2018-01-05
Add Indonesian!

jfly:
Thanks to Hafizh Dary and WCA Delegate [Cendy Cahyo Rahmat](https://www.worldcubeassociation.org/persons/2010RAHM02) for the translation.

image


jfly:
Cendy confirmed that this looks good. Unfortunately, we changed the FMC sheet slightly before merging this up, so it's also a bit out of date. I've asked Cendy and Hafizh to update the translation, which we will take care of in a separate PR.

PR #278, opened 2018-01-06
Add Vietnamese!

jfly:
Thanks to @leduyquang753 for translating.

**I've reached out to WCA Delegate Toan Do Cong and Trịnh Nguyên Anh to confirm this translation looks good.**

image

### DIFF
--- a/webscrambles/src/i18n/vi.yml
+++ b/webscrambles/src/i18n/vi.yml
@@ -1,0 +1,50 @@
+vi:
+    fmc:
+        #original_hash: 9e60b9f
+        scrambleXofY: 'Lần giải %{scrambleIndex} trong tổng số %{scrambleCount}'
+        #original_hash: b7126cd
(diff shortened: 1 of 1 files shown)


leduyquang753:
Oh, I saw the FMC sheet and realized some of the phrases were not translated correctly.
The phrases are:
**scramble:** Should be "_Lần tráo_";
**event:** Should be "_Tối ưu hóa_";
"_lần xoay_" words should be changed to "_bước xoay_".
This because this is the first time I see the actual FMC sheet and the translation depends quite much on the context. Please correct these in the file. Thank you.

jfly:
Could you email me an updated yml file? Thanks!

leduyquang753:
Here it is:
[vi.zip](https://github.com/thewca/tnoodle/files/1608579/vi.zip)
GitHub didn't allow uploading .yml files so I had to put it into a .zip.

PR #289, opened 2018-01-08
Update hu translation.

jfly:
image


jfly:
Niki has confirmed that this looks good, including the surprising to an English speaker "1. / 3" part:

> Yes, it was intentional. We discussed internally and however 1 / 3 could be also an option, but it’s more clear this way. (1. / 3)

As such, I'm going to merge this up.

PR #297, opened 2018-01-09
Update fr translation

viroulep:


viroulep:
@jfly: what's the merging policy for this?
The tests will fail until we have all translations updated.

I've added it to the 0.13.0 milestone so that it doesn't get forgotten, but maybe I can merge it anyway?

jfly:
Yes, please merge this up when you're happy with it! As you said, travis will fail, but hopefully only due to outdated translations, and it is possible to merge up PRs in TNoodle when tests still fail (maybe a bad idea, but that's how things currently work)

PR #300, opened 2018-01-09
Update ja translation.

jfly:
image


jfly:
Yuki Tanaka is pleased with how this looks, so I'm merging it up.

PR #309, opened 2018-01-11
Adding German translation!

jfly:
Thanks to Gregor Billing for the translation.

@Laura-O, Gregor said you already approved this translation, but for the record could you comment here confirming that it looks good?

image


Laura-O:
Yep, looks good! Thanks!

PR #324, opened 2018-01-14
Update es.yml

campos20:
The ~~annoying~~ dot isn't there anymore. I'm sure this was unintentional, but we should check with the responsible for the translation.


campos20:
captura de tela de 2018-01-14 10-11-47

AlbertoPdRF:
Oh, you are right, it shouldn't be there! Thank you for noticing :)

moralsh:
Yeah, _mea culpa_, thanks for noticing!

looks perfect now :)

PR #356, opened 2018-03-26
Update Korean translation

jonatanklosko:
An update from Ilkyoo =)


jonatanklosko:
Ilkyoo double-checked this, so I'm merging this up!

PR #362, opened 2018-06-10
Force users to input number for mbld

### LINKED ISSUE #361: MBLD input
I personally feel like the MBLD input in the Login with WCA flow does not stand out enough. Maybe I'm a bad user, but I'm so used to the legacy UI's "Filename", "Password", "Scramble!" feel that I think when I saw this mbld input I assumed it wanted a file name and I was happy with the default.

I was at a competition where a competitor attempted 29 cubes rather than the current default of 28. This meant I had to do some manual json editing and it was a tad painful.

@jfly and I discussed several ways we could go about this. One way is that we could increase the default number by a multiple of 7 as people get better at mbld. I think another fun way could be to look at the current mbld record and get the next highest multiple of 7.

FatBoyXPC:
@jfly and I paired up on this. This PR removed the default number of 28 and forces a user to input the number of puzzles for the attempt, and the generate scrambles button stays disabled until a number is entered.

We then decided to add a button to suggest a number to the user. We get the 333mbf WR and calculate the attempts, then round up to the nearest multiple of 7 (that's currently how many scrambles fit on a page).

This fixes #361

### DIFF
--- a/tnoodle-ui/README.md
+++ b/tnoodle-ui/README.md
@@ -12,7 +12,7 @@ This project was bootstrapped with [Create React App](https://github.com/faceboo
 
 ## Connecting to staging WCA server
 
-Set the `staging` url parameter. For example, try going to http://localhost:3001/staging=true
+Set the `staging` url parameter. For example, try going to http://localhost:3001?staging=true
 
 ## Connecting to a development WCA server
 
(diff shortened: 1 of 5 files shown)


lgarron:
This looks snazzy, but I would also recommend erring on the side of simplicity. Could we just generate 60 by default?

jfly:
@lgarron, thanks for the feedback! Let's continue this discussion over on the "Proposed change to how TNoodle asks for multibld attempts" email thread we've got going already

PR #367, opened 2018-10-07
Make TNoodle compile with java 10

timreyn:
This fixes two issues:
-Our import of winstone uses a NotImplementedException that's a private API and not exposed in java 10.
-The Float() constructors are deprecated, with Float.valueOf() as the suggested replacement.


jfly:
Nice! The build looks happy, so I'm going to merge this. Thanks, @timreyn!

PR #368, opened 2018-10-07
Fix lack of thread safety in SkewbSolver

timreyn:
Instead of having solution_length be a class variable, which leads to thread safety issues causing too-short scrambles when Skewb scrambles are generated in multiple threads.

Multiple threads can simultaneously update solution_length, which can result in shorter-than-expected scrambles being returned by the SkewbSolver.  Instead of returning an 11-move scramble solving the random position as expected, it may return an N-move substring of that scramble.

Additionally, add asserts to make bad scramble lengths an error for 2x2, Skewb, and Pyraminx.  This makes the existing tests fail on my laptop.

### DIFF
--- a/scrambles/src/puzzle/PyraminxPuzzle.java
+++ b/scrambles/src/puzzle/PyraminxPuzzle.java
@@ -41,6 +41,7 @@ public class PyraminxPuzzle extends Puzzle {
     public PuzzleStateAndGenerator generateRandomMoves(Random r) {
         PyraminxSolverState state = pyraminxSolver.randomState(r);
         String scramble = pyraminxSolver.generateExactly(state, MIN_SCRAMBLE_LENGTH, false);
+        azzertEquals(scramble.split(" ").length, MIN_SCRAMBLE_LENGTH + state.unsolvedTips());
 
         PuzzleState pState;
         try {
--- a/scrambles/src/puzzle/SkewbPuzzle.java
+++ b/scrambles/src/puzzle/SkewbPuzzle.java
@@ -1,6 +1,7 @@
 package puzzle;
 
 import static net.gnehzr.tnoodle.utils.GwtSafeUtils.azzert;
+import static net.gnehzr.tnoodle.utils.GwtSafeUtils.azzertEquals;
 
 import net.gnehzr.tnoodle.svglite.Color;
 import net.gnehzr.tnoodle.svglite.Svg;
@@ -42,6 +43,7 @@ public class SkewbPuzzle extends Puzzle {
     public PuzzleStateAndGenerator generateRandomMoves(Random r) {
         SkewbSolverState state = skewbSolver.randomState(r);
         String scramble = skewbSolver.generateExactly(state, MIN_SCRAMBLE_LENGTH, r);
+        azzertEquals(scramble.split(" ").length, MIN_SCRAMBLE_LENGTH);
 
         PuzzleState pState;
         try {
--- a/scrambles/src/puzzle/SkewbSolver.java
+++ b/scrambles/src/puzzle/SkewbSolver.java
@@ -180,14 +180,16 @@ public class SkewbSolver {
         }
     }
 
-    protected boolean search(int depth, int perm, int twst, int maxl, int lm, int[] sol, Random randomizeMoves) {
+    protected int search(int depth, int perm, int twst, int maxl, int lm, int[] sol, Random randomizeMoves) {
         if (maxl == 0) {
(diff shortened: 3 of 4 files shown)


jfly:
Thanks, Tim! Very clean implementation. Thanks for going to the extra effort of adding the asserts so our tests fail if we ever regress on this.

jfly:
For posterity, I've written up an explanation of this bug as I understand it. @timreyn, please chime in if this does not fit with your understanding!

<hr>

Here's a rough outline of how TNoodle generates skewb scrambles (code here:  https://github.com/thewca/tnoodle/blob/v0.13.4/scrambles/src/net/gnehzr/tnoodle/scrambles/Puzzle.java#L99-L105)

1. Pick a random state, R.
2. Solve random state R in exactly 11 moves. Call that generator S (aka: the scramble. if you apply S to a solved skewb, you (should) end up at R).
3. Check if R can be solved in <= 6 moves. If it can, then go back to step 1. [Else](https://www.smbc-comics.com/comic/the-program), we're happy with this state, so return S!

Steps 1 and 3 are fine in TNoodle 0.13.4. The bug Tim found is in step 2, specifically that:

S might get truncated during step 2 (specifically if [solution_length changes to -1 during this for loop](https://github.com/thewca/tnoodle/blob/v0.13.4/scrambles/src/puzzle/SkewbSolver.java#L257)).

Unfortunately, step 3 doesn't actually check if S generates R, so when it applies the filtering check on R, it's not necessarily verifying anything about S. It might be good to add that assertion, but it might have a measurable effect on performance.

jfly:
Also dumping some useful historical information from a private email thread:

<hr>
@timreyn said:

Jeremy and I have been investigating this most of today, and we diagnosed and fixed the issue.

I confirmed my theory: the skewb solver was returning substrings of the intended scrambles in some situations.  I made https://github.com/thewca/tnoodle/pull/368 to resolve this issue, and adding a check to make TNoodle (and tests) fail if something similar happens in the future.  Jeremy's merging it, and will release a new TNoodle when that's done.

We confirmed that none of the scrambles could be solved in < 7 turns, so none of the scrambles themselves were invalid.

Jeremy optimally solved all of them, the results are:

7 turns: 51
8 turns: 84
9 turns: 61

Based on Jaap's page (https://www.jaapsch.net/puzzles/skewb.htm) the expected distribution for 196 scrambles, subject to all of them being >= 7 turns, would be

7 turns: 20
8 turns: 78
9 turns: 93
10 turns: 5

Jeremy's running more analysis on all scrambles in the database.

<hr>
@jfly said:

Great work, Tim!

I've completed an analysis of optimal solutions of all 56642 skewb scrambles in our database. Here's a link to the script I used in case we want to do this analysis again: https://github.com/thewca/tnoodle/commit/70439d80c6b8100293ec2a55fb80ef6a9f7b5fc8, or against 2x2x2 or Pyraminx)

7 turns: 5894
8 turns: 22552
9 turns: 26698
10 turns: 1498
11 turns: 0

The expected distribution according to Jaap's page:

7 turns: 5669
8 turns: 22041
9 turns: 26184
10 turns: 1457
11 turns: 1.6

So this looks pretty good to me! I wonder when we'll get our first 11 mover in comp...

timreyn:
@jfly : Mostly correct!  The one fix is that, after (2), we don't use R anymore.  Instead, we generate the position that S generates, call it R' (which should be the same as R, but isn't because of this bug), and check that R' can't be solved in <7 moves.

This means that it wasn't luck; all of the scrambles were verified to not have a <7 move solution.

jfly:
Ah! I had forgotten about this code: https://github.com/thewca/tnoodle/blob/v0.13.4/scrambles/src/puzzle/SkewbPuzzle.java#L43-L53, where we take the (possibly truncated) scramble, and apply it to the solved state.

That's good, I'm glad TNoodle is architect-ed like that.

timreyn:
Here's my version of the summary, for posterity:

In the old code, when this bug happened, we would do the following:

- generate a uniformly random position R
- generate an 11-move scramble S that generates R
- drop 1 or more moves from the end of S [this is the part that's not supposed to happen!]  Call it S'.
- check that S' can't be solved in <7 moves
- output S'

This is not a purely cosmetic change, it changes the distribution of scrambles that were generated.  As we saw, the distribution of scramble depths changed.

However, no scrambles could possibly have been generated which could be solved in <7 turns.  Every scramble that was printed on the scramble sheet was checked to not be solvable in <7 moves.  Jeremy and I also verified this fact.

PR #369, opened 2018-10-07
Get rid of some python warnings from tmt.

jfly:
> ./tmt:4: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
>   import sys, os, imp, re, subprocess, shutil, shlex
> ./tmt:661: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
>   class CaseInsensitiveDict(collections.Mapping):

### DIFF
--- a/tmt
+++ b/tmt
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 from os.path import exists, dirname, join, isdir, split, abspath, relpath, basename
-import sys, os, imp, re, subprocess, shutil, shlex
+import sys, os, importlib, re, subprocess, shutil, shlex
 import zipfile
 import tempfile
 import time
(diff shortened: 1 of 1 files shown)


jfly:
Blargh, this doesn't work on older versions of python 3 (such as 3.4.3, which Travis is using). I'll leave this unmerged for now, but keep it around so we have a fix once Python 3.8 comes out). Hopefully Travis will have upgraded their version of python by then.

jfly:
I accidentally merged this, and I can't find a way to reopen this PR, so I've created https://github.com/thewca/tnoodle/pull/370 as a replacement.

PR #370, opened 2018-10-07
Get rid of some python warnings from tmt.

jfly:
> ./tmt:4: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
>   import sys, os, imp, re, subprocess, shutil, shlex
> ./tmt:661: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
>   class CaseInsensitiveDict(collections.Mapping):

### DIFF
--- a/tmt
+++ b/tmt
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 
 from os.path import exists, dirname, join, isdir, split, abspath, relpath, basename
-import sys, os, imp, re, subprocess, shutil, shlex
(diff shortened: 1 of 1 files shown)


jfly:
I accidentally merged up https://github.com/thewca/tnoodle/pull/369 and didn't mean to. I've reverted that commit and opened this PR as a replacement.

jfly:
@suushiemaniac is working to replace `tmt`, so I think we can let this PR go

PR #371, opened 2018-10-09
Update outdated info in `readme-scrambles.md`.

lgarron:
- Add Skewb.
- Mention that the requirements *are* in the Regs.
- Change two phrases to "scramble sequence" for consistency with official
  terminology.


jfly:
Where are we with the tnoodle release? Do you want us to spin out a new version of tnoodle with this change?

Issue #382, opened 2018-11-19
Generate output that helps restrict scramble access


lgarron:
An easy first step:

Put all the current TNoodle output in one subfolder, and make another folder called `SEND THIS FOLDER TO THE SCRAMBLE TABLES IF YOU'RE USING COMPUTERS AT A COMPETITION` (exact name TBD), with a PDF per scramble set encrypted using the same password as the main `.zip`.

A slightly fancier approach:

> A simple strawperson would be: single page web app capable of displaying a scramble set, stores encrypted scramble with a different key for every set, Delegate has an app that allows releasing a given set. Maybe a password entry fallback on the scramble computer client-side just so that in the worst case it’s no harder to use than .zip files.
> https://worldcubeassociation.slack.com/archives/C0GHEEJ3S/p1542654595093900

lgarron:
Also note: the problem of *releasing* access to scrambles is easy, but preventing scrambles from being accessible after is more difficult. I think we can mostly solve that issue having the web app version do a live check that the group is current, and insist on switching as soon as the Delegate marks the next round is ready.

Even better would be to integrate the round advancement into something that makes it hard to forget, such as an API that updates the projector (if the comp has one) or a screen that displays competitors to call up.

viroulep:
Maybe just renaming folders would work just as good for a first action.
Eg: rename "pdf" to "all_scrambles_set_pdfs", and put the big pdf in a subfolder "printable_scrambles".

> A slightly fancier approach: [...]

This is indeed a bit fancy, and probably overkill for most competitions.
I like the idea of something appropriate to release only one scramble set at a time, operated only by the Delegate(s) (basically a friendly tool to safely operate the encrypted zip).

suushiemaniac:
Please do not name the nested folder anything vaguely related to "Send" like in your OP suggestion.
Sending implies unzipping, which in turn nullifies the desired effect of the encryption

lgarron:
> Please do not name the nested folder anything vaguely related to "Send" like in your OP suggestion.
> Sending implies unzipping, which in turn nullifies the desired effect of the encryption

In my suggestion, each PDF inside that folder is encrypted one more time (each by itself).

bugybunny:
How is it done now? I’m not a delegate and don’t know much about it.

TNoodle generates an encrypted .zip and PDFs are just inside this archive without any subfolder or structures and the delegate needs to open the correct scrambles by just looking at the filename?

jfly:
[...]

PR #383, opened 2018-11-30
Introduce encrypted scramble set PDFs, and reorganize `.zip` structure.

lgarron:
New top-level folder structure:
- `computer-display`: One PDF for each group, encrypted using the same password as the main `.zip` file.
- `printing`: All the old unencrypted PDFs.
- `interchange`: All the old non-PDF formats.

---------------

I realize there's some code redundancy in this PR, but I've tried to make this PR as clearly "correct" as possible without fancy tricks or refactoring.

It's also not a full solution to #382, but gets us a significant part of the way there.

Possible improvements:
- Encrypt each PDF using a separate passcode, and give the Delegate a file with the passcodes so that they can hand out decryption powers more granularly.
- Also include a .zip of the computer-display folder, for easy distribution. (Sharing a folder of files without re-zipping is not simple.)

### DIFF
--- a/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
+++ b/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
@@ -64,10 +64,12 @@ import java.io.InputStreamReader;
 import java.io.StringReader;
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
+import java.security.SecureRandom;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.Map;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
@@ -251,7 +253,7 @@ class ScrambleRequest {
         return scrambleRequests;
     }
 
-    private static PdfReader createPdf(String globalTitle, Date creationDate, ScrambleRequest scrambleRequest, Locale locale) throws DocumentException, IOException {
+    private static ByteArrayOutputStream createPdf(String globalTitle, Date creationDate, ScrambleRequest scrambleRequest, Locale locale, String password) throws DocumentException, IOException {
         // 333mbf is handled pretty specially: each "scramble" is actually a newline separated
         // list of 333ni scrambles.
         // If we detect that we're dealing with 333mbf, then we will generate 1 sheet per attempt,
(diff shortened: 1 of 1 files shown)


lgarron:
Sample `.zip` file:
[Test Competition.zip](https://github.com/thewca/tnoodle/files/2632977/Test.Competition.zip)

lgarron:
Sample `.zip` with individual passcodes per scramble set PDF: 
[Test Competition.zip](https://github.com/thewca/tnoodle/files/2633126/Test.Competition.zip)

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -357,8 +370,7 @@ private static PdfReader createPdf(String globalTitle, Date creationDate, Scramb
         doc.close();
 
         // TODO - is there a better way to convert from a PdfWriter to a PdfReader?
```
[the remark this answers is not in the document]
campos20: I agree we should split createPdf. I was also considering a different class for fmc and we can call it with addFmc(). Fmc sheet is way different from regylar scrambles. It looks messy to code together.

lgarron:
@jfly, could you review again?

Here's the latest structure:
screen shot 2018-12-02 at 01 26 44

lgarron:
`.zip`: [Scrambles for Test Competition.zip](https://github.com/thewca/tnoodle/files/2636817/Scrambles.for.Test.Competition.zip)

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -276,21 +281,26 @@ private static PdfReader createPdf(String globalTitle, Date creationDate, Scramb
                 attemptRequest.event = "333bf";
                 attemptRequest.colorScheme = scrambleRequest.colorScheme;
 
-                PdfReader pdfReader = createPdf(globalTitle, creationDate, attemptRequest, locale);
+                PdfReader pdfReader = new PdfReader(createPdf(globalTitle, creationDate, attemptRequest, locale, null).toByteArray());
```
jfly: Could you put a quick comment here about why password is null?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -1312,12 +1338,23 @@ public static ByteArrayOutputStream requestsToZip(ServletContext context, String
         ZipOutputStream zipOut = new ZipOutputStream(baosZip);
         HashMap<String, Boolean> seenTitles = new HashMap<String, Boolean>();
 
+        // Computer display zip
```
jfly: Could you add a quick comment here about the goals of this zip file? I think it's easy to get confused here about if zip files are created just to easily hold files vs to encrypt stuff. I'm thinking something like:

> This zip file is intentionally *not* password protected, it's just an easy way to distribute a collection of files. Any files we put in this zip must be encrypted themselves.

(feel free to reword, but I hope the gist of what I'm trying to say makes sense)

lgarron:
This change has been approved by the Board for a TNoodle release!

jfly:
LGTM!

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
lgarron: Done!

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
lgarron: Done, thanks!

PR #390, opened 2019-01-30
Add minimum puzzles for MBLD

campos20:
This will prevent input mistakes.

### DIFF
--- a/tnoodle-ui/src/ManageCompetition.js
+++ b/tnoodle-ui/src/ManageCompetition.js
@@ -126,9 +126,10 @@ class ManageCompetition extends Component {
       </div>;
     } else {
       let disabled, title;
-      if(hasMbld && !puzzlesPer333mbfAttempt) {
+      // perhaps we should set also an upper bound to avoid numbers like 2e3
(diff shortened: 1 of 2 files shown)


tnoodle-ui/src/ManageCompetition.js
```diff
@@ -126,9 +126,11 @@ class ManageCompetition extends Component {
       </div>;
     } else {
       let disabled, title;
-      if(hasMbld && !puzzlesPer333mbfAttempt) {
+      const WCA_MINIMUM_MBLD_PER_ATTEMPT = 2;
+      // perhaps we should set also an upper bound to avoid numbers like 2e3
+      if(hasMbld && (!puzzlesPer333mbfAttempt || puzzlesPer333mbfAttempt < 2)) {
```
jfly: Can we change this to reuse the variable you just declared?

```suggestion
      if(hasMbld && (!puzzlesPer333mbfAttempt || puzzlesPer333mbfAttempt < WCA_MINIMUM_MBLD_PER_ATTEMPT)) {
```

PR #393, opened 2019-02-11
Fix style of FMC cutout sheet. This also fixes #394

### LINKED ISSUE #394: Make FM cutout sheet scrambles bigger
Here's how it looks now:
image

There's almost enough space for a 9th row, but I think it would be better to make the images bigger using the height available. This would probably mean splitting the scramble sequence in two lines, but that's already the case for all other events, so not a big issue, I believe.

campos20:
We used to end the competition name with a colon

captura de tela de 2019-02-11 14-55-09

Eventually, there was a colon mixed with a hyphen.

captura de tela de 2019-02-11 14-55-54

Colon has been removed

Updates with the latest commit (drop the use of several arbitrary number and better space use).

cutout

cutout2

### DIFF
--- a/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
+++ b/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
@@ -877,18 +877,15 @@ class ScrambleRequest {
         int height = top - bottom;
         int width = right - left;
 
-        float fontSize = 12;
-        int padding = 90;
-        int marginBottom = 10;
-        int offsetTop = 27;
-
-        int availableScrambleSpace = width - padding;
-        int scrambleFontSize = 20;
-        int titleFontSize = 15;
+        int spaceScrambleImage = 5; // scramble image won't touch the scramble
+        int scrambleImagePadding = 8; // scramble image won't touch the dashed lines
+        int fontSize = 20;
+        
+        final int scramblesPerSheet = 8;
+        int availableScrambleHeight = height/scramblesPerSheet;
 
         int availableScrambleWidth = (int) (width * .45);
-        int availableScrambleHeight = height - (int) (height * .77) - 90;
-        Dimension dim = scrambleRequest.scrambler.getPreferredSize(availableScrambleWidth - 8, availableScrambleHeight - 8);
+        Dimension dim = scrambleRequest.scrambler.getPreferredSize(availableScrambleWidth, availableScrambleHeight - 2*scrambleImagePadding);
         PdfTemplate tp = cb.createTemplate(dim.width, dim.height);
         Graphics2D g2 = new PdfGraphics2D(tp, dim.width, dim.height, new DefaultFontMapper());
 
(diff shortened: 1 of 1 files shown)


webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -1387,6 +1385,8 @@ public static ByteArrayOutputStream requestsToZip(ServletContext context, String
                     doc.addTitle(globalTitle);
                 }
 
+                // perhaps we should use i18n on the cutout sheet (scramblexofY)
```
jfly: Could you file an issue for this instead? I don't like tracking work in code comments because you can't have a conversation.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -914,18 +914,16 @@ private static void addFmcScrambleCutoutSheet(PdfWriter docWriter, Document doc,
...
-                title = globalTitle + " - " + scrambleRequest.title + ":";
+                title += " - Scramble " + (index + 1) + " of " + scrambleRequest.scrambles.length;
             }
 
             fontSize = fitText(new Font(bf), title, new Rectangle(availableScrambleSpace, 100), scrambleFontSize, false, 1f);
 
             cb.setFontAndSize(bf, fontSize);
             cb.showTextAligned(PdfContentByte.ALIGN_LEFT, title, left, top - offsetTop, 0);
-            cb.showTextAligned(PdfContentByte.ALIGN_LEFT, globalTitle + " - " + scrambleRequest.title + ":", left, top - offsetTop, 0);
+            cb.showTextAligned(PdfContentByte.ALIGN_LEFT, globalTitle + " - " + scrambleRequest.title, left, top - offsetTop, 0);
```
jfly: I'm having trouble understanding how this line of code interacts with the previous line of code. It looks like they're drawing to the exact same coordinates... why aren't they overlapping?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: Sure!

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
campos20: These bugs are currently been used on TNoodle. I realized it on my [last competition](https://www.worldcubeassociation.org/competitions/FMCBrasil2019).

I'll replace these line with fixAndShowText. This sould save a few lines and make it more clear.

campos20: You were correct. There was 2 titles shown at the same spot.
[This](https://github.com/thewca/tnoodle/blob/c798b44b78757bf39516516b32d06c837d256661/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L927) and [this](https://github.com/thewca/tnoodle/blob/c798b44b78757bf39516516b32d06c837d256661/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L928).

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java

campos20:
This has become related with #394 

captura de tela de 2019-02-24 20-01-57

@pedrosino if this it's ok, I can put this PR to close #394

pedrosino:
I still think there's some wasted space in the sheet.
Can you make a side-to-side comparison?

campos20:
Sure!

captura de tela de 2019-02-26 13-50-25

Proposed / current

This can be easily adjusted, but I'm afraid we run into some not printable area thing.

pedrosino:
Your version of current looks different from mine at #394

I'm fine with this change as it is. Perhaps moving the scramble a bit down would help not to miss the final F move.

campos20:
> Your version of current looks different from mine at #394

I regenerated and I got much more space now. Perhaps this is because the current version [calculates](https://github.com/thewca/tnoodle/blob/c798b44b78757bf39516516b32d06c837d256661/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L924) the fontsize, [use it](https://github.com/thewca/tnoodle/blob/c798b44b78757bf39516516b32d06c837d256661/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L930) to generate more space. This is now fixed since the [available space is calculated](https://github.com/thewca/tnoodle/blob/9871dc2eb1abdaedb74a6e0341620284dade14f4/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L885) and then [we fit](https://github.com/thewca/tnoodle/blob/9871dc2eb1abdaedb74a6e0341620284dade14f4/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L921) the image and scramble on each available space.

> I'm fine with this change as it is. Perhaps moving the scramble a bit down would help not to miss the final F move.

Ok, I'll put the PR to close this. The move won't be cancelled, neither the competition names. Font sizes [are calculated](https://github.com/thewca/tnoodle/blob/9871dc2eb1abdaedb74a6e0341620284dade14f4/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L922) (after the calculation of the available space).

When I mentioned not printable area it was considering the margins (left and right). Perhaps if we go too far left, we may run into printers issue.

PR #395, opened 2019-02-21
Add ordered scrambles. This fixes #338.

### LINKED ISSUE #338: Order scrambles by schedule
Iskender commented that he could arrange the round order in the old ui to save the time needed to sort the printed sheets.
Once we'll have the schedule in the WCIF, we'll be able to add this neat feature without any additional input from the Delegate!

campos20:
Zip's root folder doesn't change. Inside /Printing, there's a new folder called Ordered Scrambles.

ordered1

Iff the competition has multiple venues, one folder for each is created (in the example, a venue was called Temple1/Temple2).

ordered2

Iff the competition has multiple days, one folder for each day is created (in the example, day 2 has no events).

ordered3

Finally, each room has its own pdf, whose rounds order must match the competition schedule. If there's just 1 room available for that venue, the room's name is not stamped (the same goes for venue and day).

ordered4

If, for a competition, there's just 1 venue, 1 room and 1 day, the unique created pdf should be /Printing/Ordered Scrambles/Ordered Scrambles.pdf

The main idea here is that the delegate can open a file, print it and send to a room.

Room's and venue's name only consider the part inside brackets (for markdown, if available), each slash is replaced with a space (to prevent creation of unwanted folders) and multiple whitespaces are replaced with single spaces.

Legacy should stay as is.


campos20:
This is currently generating scrambles for Welcometo2019 as if it was a 1 day competition.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
```diff
@@ -0,0 +1,276 @@
...
+
+    // In case venue or room is using markdown
+    private static String parseMarkdown(String s) {
+        if (s.indexOf('[') >= 0 && s.indexOf(']') >= 0) {
+            s = s.substring(s.indexOf('[')+1, s.indexOf(']'));
+        }
+        return s;
+    }
+
+    private static long dayDifference(Date date1, Date date2) {
```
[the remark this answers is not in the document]
campos20: Fixed

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleViewHandler.java
```diff
@@ -223,12 +223,12 @@ protected void wrappedService(HttpServletRequest request, HttpServletResponse re
             }
             query = parseQuery(body.toString());
 
-            String json = query.get("sheets");
+            String json = query.get("sheets"); // all the scrambles
```
jfly: Why do we need to pull out this json string and pass it down? Can we instead use the parsed `scrambleRequests` variable? I see that we currently are just reparsing this in OrderedScrambles.java, and that feels redundant to me.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleViewHandler.java
```diff
@@ -243,8 +243,11 @@ protected void wrappedService(HttpServletRequest request, HttpServletResponse re
 
                 sendBytes(request, response, totalPdfOutput, "application/pdf");
             } else if (extension.equals("zip")) {
+                String schedule = query.remove("schedule");
```
jfly: Similar question: could we instead parse the JSON here, rather than passing the string all the way down? That feels more correct to me.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -146,14 +147,16 @@ private static BaseFont getFontForLocale(Locale locale) {
...
 
-
     public String[] scrambles;
     public String[] extraScrambles = new String[0];
     public Puzzle scrambler;
     public int copies;
     public String title;
     public boolean fmc;
     public HashMap<String, Color> colorScheme;
+    public int totalAttempt;
```
jfly: Could you add a comment explaining what values this can take on? In particular, I think you set this to -1 when it's unset? It would be nice to explain that somewhere.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
```diff
@@ -0,0 +1,245 @@
...
+public class OrderedScrambles {
+    
+    // TODO we are using Joda-Time for dates, via its .jar file
+    // We can update java to java8 and use java.time, which is the same class, as of JSR-310
+    // We can dismiss joda-time-2.10.1.jar from /lib
+
+    private static final String wcifIgnorableKey = "other";
+
+    public static void generateOrderedScrambles(String globalTitle, Date generationDate, ZipOutputStream zipOut, ZipParameters parameters, String scheduleJsonStringfied, String sheet) throws DocumentException, IOException, ZipException {
+        JsonObject scheduleJson = new JsonParser().parse(scheduleJsonStringfied).getAsJsonObject();
```
jfly: I think this would be cleaner if we instead declare our own java class for the WCIF schedule, and let GSON parse the json into that for us. I think that will make some stuff nicer:

- integers will get convert to actual java integers (I see you're calling `Integer.parseInt` below)
- you should be able to access attributes via `.foo` rather than `.get` or `getAsJsonArray`

I think you can see an example of how to do this with how we parse the `ScrambleRequest` class.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
```diff
@@ -0,0 +1,245 @@
...
+        if (s.indexOf('[') >= 0 && s.indexOf(']') >= 0) {
+            s = s.substring(s.indexOf('[')+1, s.indexOf(']'));
+        }
+        return s;
+    }
+
+    private static String getEarlierActivityString(JsonObject scheduleJson) {
+
+        DateTime date = null;
+        String out = "";
```
jfly: nit: I think it's cleaner to set this to `null` initially (don't forget to update the check down on line 242 as well)

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
```diff
@@ -0,0 +1,245 @@
...
+                }
+            }
+        }
+    }
+
+    private static boolean compareFirstCharToNumber(String letter, int number) {
+        return letter.charAt(0)-'A' == number-1;
+    }
+
+    private static String removeQuotation(String s) {
```
jfly: This feels weird to me. Why are quotation marks showing up in these strings?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
```diff
@@ -0,0 +1,245 @@
...
+                        ByteArrayOutputStream baos = ScrambleRequest.requestsToPdf(globalTitle, generationDate, scrambleRequests, null);
+                        zipOut.write(baos.toByteArray());
+                        zipOut.closeEntry();
+                    }
+                }
+            }
+        }
+    }
+
+    private static boolean compareFirstCharToNumber(String letter, int number) {
```
jfly: Does this handle the case where we have more than 26 groups? At that point, TNoodle goes from group `Z` to group `AA`, I think.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
```diff
@@ -0,0 +1,245 @@
...
+                                group = Integer.parseInt(item.substring(1));
+                            } else if (item.charAt(0) == 'a') {
+                                attempt = Integer.parseInt(item.substring(1));
+                            }
+                        }
+
+                        ScrambleRequest[] scrambleRequests = GSON.fromJson(sheet, ScrambleRequest[].class);
+                        
+                        // first, we add all requests whose events equals what we need
+                        ArrayList<ScrambleRequest> scrambleRequestTemp = new ArrayList<ScrambleRequest>();
```
jfly: nit: It might be cleaner/easier to first build a `HashMap` of `String eventId` -> `ScrambleRequest[]` and then do a lookup into it at this point.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
```diff
@@ -0,0 +1,245 @@
...
+
+                        // then, we start removing
+                        if (round > 0) {
+                            ArrayList<ScrambleRequest> temp = new ArrayList<ScrambleRequest>();
+                            for (ScrambleRequest scrambleRequest : scrambleRequestTemp) {
+                                if (scrambleRequest.round == round) {
+                                    temp.add(scrambleRequest);
+                                }
+                            }
+                            scrambleRequestTemp = new ArrayList<ScrambleRequest>(temp);
```
jfly: There are enough variables with the name "temp" in them that I'm getting confused. Can you think of a better name for `scrambleRequestTemp`?

update: After reading all this code, I think the high level idea here is that you have to search over the existing scramble requests to find the ones that match this wcif activity? It might be cleaner to pull that out into its own function... something that takes the wcif activity and the list of all scramble requests and returns the scramble requests that match that activity?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleViewHandler.java
campos20: I just added the //comment while I was trying to understand the flux from front-end to the back-end. I'd rather not to change that.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
campos20: Fixed

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
```diff
@@ -0,0 +1,245 @@
...
+                            if (item.charAt(0) == 'r') {
+                                round = Integer.parseInt(item.substring(1));
+                            } else if (item.charAt(0) == 'g') {
+                                group = Integer.parseInt(item.substring(1));
+                            } else if (item.charAt(0) == 'a') {
+                                attempt = Integer.parseInt(item.substring(1));
+                            }
+                        }
+
+                        ScrambleRequest[] scrambleRequests = GSON.fromJson(sheet, ScrambleRequest[].class);
```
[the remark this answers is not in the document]
campos20: Fixed, maybe

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleViewHandler.java
campos20: Fixed, I think

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
campos20: Fixed. No removeQuotation is used anymore.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
campos20: Fixed

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleViewHandler.java
campos20: Nah, fixed.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
campos20: Done.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
campos20: Since this takes writing an iteration anyways and it doesn't change too much on computation, unless you have strong feeling about this, I'd rather not to use for this (moved to the new WCIFHandler class).

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
campos20: Done. New WCIFHandler class.

.classpath
```diff
@@ -0,0 +1,111 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jfly: I see you've added a `.classpath` and `.project` file here. Could you explain what these are for? My idea with TNoodle was that each folder in TNoodle would be a full fledged eclipse project. See https://github.com/jfly/tnoodle/blob/7ed429c25795f830cd2f8333c9609b758f6acaed/README.md#notes for more information. I don't think that would play nice with these top level `.classpath` and `.project` files you've added here.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
```diff
@@ -0,0 +1,118 @@
...
+import net.lingala.zip4j.io.ZipOutputStream;
+import net.lingala.zip4j.model.ZipParameters;
+
+import com.google.gson.JsonElement;
+
+import com.itextpdf.text.DocumentException;
+
+public class OrderedScrambles {
+    
+    // TODO we are using Joda-Time for dates, via its .jar file
```
jfly: nit: Could you file a github issue for this and update this comment to link to it?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
```diff
@@ -0,0 +1,118 @@
...
+
+                for (int index = 0; index<dayList.size(); index++) {
+                    if (scrambleRequestListByDay.get(index).size()>0) {
+                        
+                        Collections.sort(scrambleRequestListByDay.get(index));
+                        
+                        String pdfFileName = "Printing/Ordered Scrambles/";
+
+                        if (hasMultipleVenues) {
+                            pdfFileName += venueName.replace('/', ' ')+"/";
```
jfly: Maybe use `toFileSafeString` here instead? https://github.com/jfly/tnoodle/blob/7ed429c25795f830cd2f8333c9609b758f6acaed/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L1303-L1310

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
```diff
@@ -0,0 +1,118 @@
...
+                        if (hasMultipleDays || dayList.size() > 1) { // Double check, just in case.
+                            pdfFileName += "Day "+dayList.get(index)+"/";
+                        }
+
+                        pdfFileName += "Ordered Scrambles";
+                        
+                        // In addition to different folders, we stamp venue, day and room in the PDF's name
+                        // to prevent different files with the same name.
+                        if (hasMultipleVenues) {
+                            pdfFileName += " - " + venueName.replace('/', ' ');
```
jfly: Maybe use `toFileSafeString` here instead? https://github.com/jfly/tnoodle/blob/7ed429c25795f830cd2f8333c9609b758f6acaed/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L1303-L1310

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
```diff
@@ -0,0 +1,118 @@
...
+                        }
+                        if (hasMultipleDays || dayList.size() > 1) {
+                            pdfFileName += " - Day "+dayList.get(index);
+                        }
+                        
+                        if (hasMultipleRooms) {
+                            pdfFileName += " - "+roomName.replace('/', ' ');
```
jfly: Maybe use `toFileSafeString` here instead? https://github.com/jfly/tnoodle/blob/7ed429c25795f830cd2f8333c9609b758f6acaed/webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java#L1303-L1310

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
```diff
@@ -0,0 +1,118 @@
...
+                        }
+                        pdfFileName += ".pdf";
+                        
+                        // removing slashes may add double+ extra spaces
+                        pdfFileName = pdfFileName.replaceAll("\\s+", " ");
```
jfly: (related to all my comments above about using `toFileSafeString`) This feels like reasonable logic to me. If you want to keep it around, could you add it to `toFileSafeString` instead?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleHandler.java
```diff
@@ -73,13 +72,11 @@ protected void wrappedService(HttpServletRequest request, HttpServletResponse re
 
                 sendBytes(request, response, totalPdfOutput, "application/pdf");
             } else if(ext.equals("zip")) {
-                ScrambleRequest[] scrambleRequests = ScrambleRequest.parseScrambleRequests(query, seed);
-                ByteArrayOutputStream baosZip = ScrambleRequest.requestsToZip(getServletContext(), globalTitle, generationDate, scrambleRequests, seed, generationUrl);
+                ByteArrayOutputStream baosZip = ScrambleRequest.requestsToZip(getServletContext(), globalTitle, generationDate, query);
```
jfly: Stuff like `query` should not escape out of `*Handler` classes. It violates the abstraction we're trying to have between "this is our API", and "the internal business logic of tnoodle".

I preferred how this was before: this file would handle extracting/parsing stuff from the query, and pass the structured data `requestsToZip` needs to work.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -153,6 +154,16 @@ public ScrambleRequest(){}
...
+    public int attempt;
+    public DateTime roundStartTime;
+    
+    // totalAttempt and attempt are useful for when we have multiple attempts split in the schedule.
+    // When not using ordered scrambles, we create an array of ScrambleRequest and
+    // the position in the array is related to the attempt, hence we can print attempt number (mainly for FMC and MBLD)
+    // and use, for example, scramble 1 or 3 (3 here is array.length)
+    // But when using ordered scrambles, each attempt is treated like an individual ScrambleRequest.
+    // We use totalAttempt as array.length and attempt as the position that scramble would be on the array.
```
jfly: Thanks for adding this comment. Could you move things around so they're visibly "attached" to `totalAttempt` and `attempt`? I'm thinking something like this:

```java
...
public HashMap<String, Color> colorScheme;

// BIG COMMENT HERE
// ...
public int totalAttempt;
public int attempt;

...
```

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -257,6 +268,9 @@ private static ByteArrayOutputStream createPdf(String globalTitle, Date creation
         // list of 333ni scrambles.
         // If we detect that we're dealing with 333mbf, then we will generate 1 sheet per attempt,
         // rather than 1 sheet per round (as we do with every other event).
+        
+        // for ordered scrambles, we recreate scrambleRequest so it contains only 1 scramble
+        // to fix this, we pass the attempt number
```
jfly: I'm not sure what "pass the attempt number" is referring to here. Could you update this comment a bit?

nit: Also, if you look at the other comments in TNoodle, you'll see that we tend to write in full sentences. Could you do that as well for consistency?

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -1346,6 +1366,7 @@ public static ByteArrayOutputStream requestsToZip(ServletContext context, String
         String safeGlobalTitle = toFileSafeString(globalTitle);
         String computerDisplayFileName = safeGlobalTitle + " - Computer Display PDFs";
 
+        ScrambleRequest[] scrambleRequests = GSON.fromJson(query.get("sheets"), ScrambleRequest[].class);
```
jfly: As I commented over in `ScrambleHandler`, I think this sort of string parsing belongs up in the handler class, and not down in this method.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleViewHandler.java
```diff
@@ -244,7 +242,8 @@ protected void wrappedService(HttpServletRequest request, HttpServletResponse re
                 sendBytes(request, response, totalPdfOutput, "application/pdf");
             } else if (extension.equals("zip")) {
                 ByteArrayOutputStream zipOutput = ScrambleRequest
-                        .requestsToZip(getServletContext(), globalTitle, generationDate, scrambleRequests, password, generationUrl);
+                        .requestsToZip(getServletContext(), globalTitle, generationDate, query);
```
jfly: Similar comment here that I think the query parsing should happen up here in the `*Handler` class, rather than down in `ScrambleRequest.requestsToZip`

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/WCIFHandler.java
```diff
@@ -0,0 +1,195 @@
...
+    
+    // Currently, we mark not cubing related activities as other-lunch or other-speech, for example.
+    // If we ever accept any other such ignorable key, it should be added here.
+    private static final String[] wcifIgnorableKeys = new String[]{"other"};
+
+    private ArrayList<ScrambleRequest> allScrambleRequests;
+    private JsonObject schedule;
+    private JsonParser parser;
+
+    public WCIFHandler(LinkedHashMap<String, String> query) {
```
jfly: I think this class should instead take a `ScrambleRequest[]` or `ArrayList<ScrambleRequest>`, and the string/query parsing should happen up in a `*Handler` class.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/WCIFHandler.java
```diff
@@ -0,0 +1,195 @@
...
+
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+
+public class WCIFHandler {
```
jfly: nit: Can we call this something other than `WCIFHandler`? In TNoodle, `*Handler` means "this is a servlet that handles api requests", which this class is not. Maybe call this `WCIFHelper` instead?

.classpath
campos20: Fixed

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
campos20: Done

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
campos20: Fixed

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
campos20: Fixed

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
campos20: Done

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/WCIFHandler.java
campos20: Renamed

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/WCIFHandler.java
campos20: Fixed

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleViewHandler.java
campos20: Fixed after our video chat

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
```diff
@@ -0,0 +1,120 @@
...
+                        
+                        Collections.sort(scrambleRequestListByDay.get(index));
+                        
+                        String pdfFileName = "Printing/Ordered Scrambles/";
+
+                        if (hasMultipleVenues) {
+                            pdfFileName += venueName+"/";
+                        }
+
+                        if (hasMultipleDays || dayList.size() > 1) { // Double check, just in case.
```
jfly: I think it makes more sense to move this logic up into the line of code where we first declare `hasMultipleDays`. That would save you from having to do the same check down on line 100.

I also think that when you move this up, you should add a better comment. This is a pretty tricky situation that you had to do a bit of investigation to understand, and I think this is a good opportunity to teach future code readers about what you learned.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleHandler.java
```diff
@@ -74,12 +74,14 @@ protected void wrappedService(HttpServletRequest request, HttpServletResponse re
                 sendBytes(request, response, totalPdfOutput, "application/pdf");
             } else if(ext.equals("zip")) {
                 ScrambleRequest[] scrambleRequests = ScrambleRequest.parseScrambleRequests(query, seed);
-                ByteArrayOutputStream baosZip = ScrambleRequest.requestsToZip(getServletContext(), globalTitle, generationDate, scrambleRequests, seed, generationUrl);
+                
+                System.out.println("Zip no handler");
```
jfly: Please remove this. If you need to print something, use the logging infrastructure.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleRequest.java
```diff
@@ -1570,6 +1590,9 @@ public static ByteArrayOutputStream requestsToZip(ServletContext context, String
         jsonObj.put("version", Utils.getProjectName() + "-" + Utils.getVersion());
         jsonObj.put("generationDate", generationDate);
         jsonObj.put("generationUrl", generationUrl);
+        if (wcifHelper != null) {
+            jsonObj.put("schedule", wcifHelper.getSchedule());
```
jfly: Out of curiosity, why are we putting the schedule in here? I think this JSON file is only read by the workbook assistant, and we don't have any plans to have it start reading the schedule.

That said, I don't think there's any downside to putting the schedule here, and it might be nice to have a snapshot of the schedule we used for generating scrambles, so I'm ok with leaving this here. Just want to understand why it's here.

campos20: It's mostly for the automated test I'm about to write. I also wanna check if the PDFs are in the correct order. It looks like we can change schedule after a competition is over, so using the online schedule might not be the best option.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/ScrambleHandler.java
campos20: Fixed, this was my super professional debugging.

webscrambles/src/net/gnehzr/tnoodle/server/webscrambles/OrderedScrambles.java
campos20: Fixed

PR #402, opened 2019-04-26
Show rounds and groups to the user. This Fixes #360

### LINKED ISSUE #360: Warning for rounds with one group
The first time I used the new workflow (Login with WCA), I had not yet changed the group information in the WCA's site / manage events page. I ended up with a single group for every round and was confused. Later, @jfly informed me what I needed to do and it made more sense!

I figure some sort of warning before generating the scrambles if every round has a single group would be a nice mechanism to remind users when they are being dumb :smile:

campos20:
Proposal.

React

### DIFF
--- a/tnoodle-ui/src/components/EventsTable.jsx
+++ b/tnoodle-ui/src/components/EventsTable.jsx
@@ -15,7 +15,7 @@ class EventsTable extends Component {
               {Array.from({ length: maxRounds }, (_, i) => {
                 return (
                   <th key={i} scope="col">
-                    Groups for Round {i + 1}
+                    Scramble Sets for Round {i + 1}
                   </th>
                 );
               })}
--- a/tnoodle-ui/src/components/EventsTable.jsx
+++ b/tnoodle-ui/src/components/EventsTable.jsx
@@ -1,0 +1,51 @@
+import React, { Component } from "react";
+import CubingIcon from "../CubingIcon";
+
+class EventsTable extends Component {
+  render() {
+    const { events } = this.props;
+    let maxRounds = Math.max.apply(
+      Math,
+      events.map(event => event.rounds.length)
+    );
+    return (
+      <div className="text-center">
+        <table className="table table-striped table-bordered table-hover table-condensed">
+          <thead className="thead-dark">
+            <tr>
+              <th scope="col">Event</th>
+              <th scope="col">Event Name</th>
+              {Array.from(Array(maxRounds), (e, i) => {
+                return (
+                  <th key={i} scope="col">
+                    Groups for Round {i + 1}
+                  </th>
+                );
+              })}
+            </tr>
+          </thead>
+          <tbody>
+            {events.map(event => (
+              <tr>
+                <td>
+                  <CubingIcon key={event.id} event={event.id} />
+                </td>
+                <td className="align-middle">{event.id}</td>
+                {event.rounds.map(round => (
+                  <td className="align-middle">{round.scrambleSetCount}</td>
+                ))}
+
+                {/* Fill empty cells with --- */}
+                {Array.from(Array(maxRounds - event.rounds.length), () => {
+                  return <td className="align-middle">-- -- --</td>;
+                })}
+              </tr>
+            ))}
+          </tbody>
+        </table>
+      </div>
+    );
+  }
+}
+
+export default EventsTable;
--- a/tnoodle-ui/src/components/EventsTable.jsx
+++ b/tnoodle-ui/src/components/EventsTable.jsx
@@ -13,8 +13,8 @@ class EventsTable extends Component {
         <table className="table table-striped table-bordered table-hover table-condensed">
           <thead className="thead-dark">
             <tr>
+              <th scope="col" />
               <th scope="col">Event</th>
-              <th scope="col">Event Name</th>
               {Array.from(Array(maxRounds), (e, i) => {
                 return (
                   <th key={i} scope="col">
--- a/tnoodle-ui/src/components/EventsTable.jsx
+++ b/tnoodle-ui/src/components/EventsTable.jsx
@@ -4,10 +4,7 @@ import CubingIcon from "../CubingIcon";
 class EventsTable extends Component {
   render() {
     const { events } = this.props;
-    let maxRounds = Math.max.apply(
-      Math,
-      events.map(event => event.rounds.length)
-    );
+    let maxRounds = Math.max(...events.map(event => event.rounds.length));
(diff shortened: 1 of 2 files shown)


campos20:
@FatBoyXPC we could use your opinion here.

tnoodle-ui/src/components/EventsTable.jsx
```diff
@@ -0,0 +1,51 @@
...
+      events.map(event => event.rounds.length)
+    );
+    return (
+      <div className="text-center">
+        <table className="table table-striped table-bordered table-hover table-condensed">
+          <thead className="thead-dark">
+            <tr>
+              <th scope="col" />
+              <th scope="col">Event</th>
+              {Array.from(Array(maxRounds), (e, i) => {
```
jfly: I'd never heard of `Array.from` until now, but from some googling (https://stackoverflow.com/questions/26052699/array-from-vs-array-prototype-map), it sounds like it's useful for dealing with "array like" objects, and is maybe a bit let performant than `Array.map`. In this situation, we're dealing with a plain array (not an array-like object), so I think it would be less surprising to use `map` here instead. I'd suggest rewriting this as:

```javascript
Array(maxRounds).map((e, i) => {....
```

tnoodle-ui/src/components/EventsTable.jsx
```diff
@@ -0,0 +1,51 @@
...
+                  </th>
+                );
+              })}
+            </tr>
+          </thead>
+          <tbody>
+            {events.map(event => (
+              <tr>
+                <td>
+                  <CubingIcon key={event.id} event={event.id} />
```
jfly: I think the `key=` is unnecessary here, as we're not in a loop at this point.

tnoodle-ui/src/components/EventsTable.jsx
```diff
@@ -0,0 +1,51 @@
...
+                  <th key={i} scope="col">
+                    Groups for Round {i + 1}
+                  </th>
+                );
+              })}
+            </tr>
+          </thead>
+          <tbody>
+            {events.map(event => (
+              <tr>
```
jfly: Does this `tr` need a `key` attribute? I'd expect you to see warnings in the console if you don't have one.

tnoodle-ui/src/components/EventsTable.jsx
```diff
@@ -0,0 +1,51 @@
...
+                <td>
+                  <CubingIcon key={event.id} event={event.id} />
+                </td>
+                <td className="align-middle">{event.id}</td>
+                {event.rounds.map(round => (
+                  <td className="align-middle">{round.scrambleSetCount}</td>
```
jfly: Does this `td` need a `key` attribute? I'd expect you to see warnings in the console if you don't have one.

tnoodle-ui/src/components/EventsTable.jsx
```diff
@@ -0,0 +1,51 @@
...
+                ))}
+
+                {/* Fill empty cells with --- */}
```
jfly: I don't think this comment matches the code (the code has more dashes?

tnoodle-ui/src/components/EventsTable.jsx
```diff
@@ -0,0 +1,51 @@
...
+                {Array.from(Array(maxRounds - event.rounds.length), () => {
```
jfly: Similar comment as above about `Array.from` vs `Array.map`.

tnoodle-ui/src/components/EventsTable.jsx
```diff
@@ -0,0 +1,51 @@
+import React, { Component } from "react";
+import CubingIcon from "../CubingIcon";
+
+class EventsTable extends Component {
+  render() {
+    const { events } = this.props;
+    let maxRounds = Math.max.apply(
```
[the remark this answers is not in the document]
jonatanklosko: I believe you meant `Math.max(...events.map(event => event.rounds.length))` =)

tnoodle-ui/src/components/EventsTable.jsx
jonatanklosko: Actually this looks like a valid use case for `Array.from` to me! See some random code:
```js
> Array(5).map(x => 10)
[ <5 empty items> ]
> Array.from(Array(5)).map(x => 10)
[ 10, 10, 10, 10, 10 ]
> Array.from({ length: 5 }, x => 10)
[ 10, 10, 10, 10, 10 ]
```
[Here](https://stackoverflow.com/a/5501711)'s a SO answer explaining that.

In this case I'd suggest this:
```js
Array.from({ length: maxRounds - event.rounds.length }, (_, i) =>
  <td key={i} className="align-middle">-- -- --</td>
);
```

tnoodle-ui/src/components/EventsTable.jsx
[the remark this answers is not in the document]
campos20: Fixed used @jonatanklosko version

tnoodle-ui/src/components/EventsTable.jsx
campos20: Changed to `[...Array(maxRounds)].map((e, i) =>`

tnoodle-ui/src/components/EventsTable.jsx
campos20: You are totally correct. Fixed.

tnoodle-ui/src/components/EventsTable.jsx
campos20: Fixed

tnoodle-ui/src/components/EventsTable.jsx
campos20: Fixed

tnoodle-ui/src/components/EventsTable.jsx
campos20: Updated

tnoodle-ui/src/components/EventsTable.jsx
campos20: Used @jonatanklosko's suggestion

campos20:
I renamed "Groups" to "Scramble Sets" since the website uses "Scramble Sets". Groups is used in WCIF though.

jfly:
> Groups is used in WCIF though

I think we discussed this over in slack, but for the record: the WCIF deals with both groups (something you may schedule to occur) and scramble sets (which can be assigned to 1 or more groups. each activity has a `scrambleSetId` to create this mapping)

PR #404, opened 2019-05-31
Bump version

viroulep:
I believe we'll need it for the upcoming release.
Also it fixes these security alerts that show up.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "tnoodlejs",
-  "version": "0.14.0",
+  "version": "0.15.0",
   "homepage": "https://github.com/thewca/tnoodle",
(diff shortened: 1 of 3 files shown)


jfly:
LGTM!

viroulep:
@jfly it seems we still have a vulnerable version of lodash in [`git-tools`](https://github.com/thewca/tnoodle/blob/master/git-tools/package-lock.json).
I'm not sure how it's built and how to upgrade that, could you look into it please?
Note that I don't think it's a big deal, as I think it's something that should be used only by developer to create a release(?).

jfly:
Oh dear. That will be going away soon with @suushiemaniac's work to port tnoodle to use gradle.

PR #409, opened 2019-06-04
So long, tmt.

### LINKED ISSUE #160: [Meta] Slim down project
As mentioned on [speedsolving](http://www.speedsolving.com/forum/showthread.php?45693-WCA-Regulations-2014-Scramble-Filtering-%28new-poll%29&p=942245&viewfull=1#post942245) TNoodle would in my opinion benefit from a slimmer overall appearance.

Here's a rough roadmap how I'd approach this. Feel free to add missing points/correct existing ones and raise any objections in the comments below.
## TODO
- [ ] Remove all projects which are not required for building TNoodle.
  - Either nuke them from earth or move them to separate repositories.
- [ ] Remove unused features from `tmt`.
  - `__pullRequest`?
- [ ] Replace most of `tmt` with a dedicated Java build system.
  - [gradle](http://www.gradle.org/) looks like a promising alternative compared to the obnoxious `ant` and `maven`.
  -  `automake`?
- [ ] Decide on the definition of `TNoodle`.
  - Is it just the scrambler or a bunch of cubing-related projects (as mentioned in the readme)?

I can (obviously) help with removing stuff, but I have absolutely no experience with the Java build ecosystem. I never used any of the alternatives posted above.

suushiemaniac:
This commit introduces the [Gradle build system](https://gradle.org/) to TNoodle. Gradle is a de-facto standard for building and maintaining complex Java projects, and it acts as a reliable and [actively maintained](https://github.com/gradle/gradle) alternative to Jeremy's own `tmt`.

This PR will **not** cover the Gradle basics. Please skim the updated `README.md` or walk through [the tutorials](https://gradle.org/guides/#getting-started). Likewise, I will not explain every Gradle term in detail below. There are glossaries online and you can always contact me if you need help :)

## Architecture
TNoodle is now organised as a *multi-project* build. Essentially, all the modules are structured as before, and I tried to stay true to the original project architecture wherever possible.

### Changed/Updated modules
- **hello-winstone**: Removed
- **quercus**: Removed
- **lib**: No external java dependencies are checked in anymore. They are all managed by Gradle through central repositories like `Maven Central` or `jcenter`.
- **tnoodlejs**: Exists as a standalone project such that the manually pasted JDK source code is not part of `scrambles` anymore
- **tnoodle-android**: Instead of `library` and `sample` there is only one merged project which remains. It's mostly identical to the previous `sample`, as publishing a dedicated `aar` was never necessary in the first place (further information down below)

### Renamings
`614 changed files` (as of writing this post) looks absolutely awful, Over half of them are actually just renames, because instead of `src` the default Gradle folder structure is `src/main/java` (or `src/main/kotlin`, respectively)

## Building / Regressions
Building the project essentially works the same way as before (multiple components that depend on each other), just that the commands which you have to type into your terminal slightly differ. You typically interact with Gradle through a so-called *Gradle Wrapper* which, well, wraps the executable bin file.

I am not providing any further instructions here on purpose, since the essentials are already outlined in the new `README.md` and everything should be self-contained.
The only thing that you **will** have to do is set up an Android SDK as per https://developer.android.com/studio (scroll all the way to the bottom and opt for "Command Line Tools only" if you want to avoid several GB of IDE overhead in downloads)

If done right, the setup should "just work" after the prerequisites have been set up. In particular, the end result in terms of deloyable JARs does not change at all with this new feature. Please be extra-careful when testing.

## On the size of this commit
I know that this pull request is massive. The problem is that I don't see an easier solution here - either you migrate the build system or you don't. The previously used `tmt` was able to invoke Gradle, but it would have been a ridiculously unnecessary pain in the butt to make Gradle invoke (and interoperate with) `tmt` just for the sake of downsizing the PR.

When touching the foundations of a complex project like this, it's only natural that a lot of commits are needed. I hope you can understand this necessity, as I am neither willing nor able to sub-divide this feature.

## Travis builds
Travis barely has to do any manual checking now, because all build checks are delegated to Gradle. I have updated the Travis config in this PR, but it's gonna be a total surprise for me to see what happens when the first build runs. My goal is <= 5 "fixes" necessary ;-)

### Logs and Tracing
Even though every check is behind Gradle now, we will still get sophisticated and dedicated log output, including traces in case something goes wrong. Gradle logs **everything**, so even if something fails in Travis we will see the erroneous output there.

## Artifacts
The following modules have been configured as deployable artifacts:
- `min2phase`
- `threephase`
- `sq12phase`
- `scrambles`
- `utils`
- `svglite`

Internally, they are all directly linked but some "external" projects like `tnoodlejs` and `tnoodle-android` depend on these artifacts instead of directly linking the source code.

In particular, these artifacts can also be **published** for public availability to any hosted Maven repository (yes, we use Gradle, but for dependency JARs the Maven standard still is the way to go). I think @jfly should be responsible for handling this (and there seems to already be an account for the old `tnoodle-android` publication which we could reuse).

### Chuang Shen's scramblers

The central `scrambles` artifact needs our custom scrambler adjustments (in particular, `min2phase`) in order to compile, so we have to publish them together with all the other artifacts.
From my understanding of GPLv3, we are allowed to publish the scramblers as well. @cs0x7f If you read this, can you please chime in and tell us whether that's okay for you?

### tnoodle-android

To this day, I fail to understand why tnoodle has been packaged as a separate `aar` artifact. **Android can consume standard JAR dependency libraries**. I suspect the reason was that this way the Swing AWT component of `min2phase` could be overridden, but I achieved the same thing in a more "proper" way so the old `library` module was removed. Android Apps can consume `scrambles`

## Testing of the new features
I need all hands that I can get to review this commit, in particular to **ensure that nothing breaks**. You should all get TNoodle to compile pretty easily (I hope...) but then please walk through the program, old and new UI alike, and click every button, call every endpoint, see that the desired PDFs are still the same.

**I did not change any scrambling code directly**. I had to touch `PuzzlePlugins` but that should be about it. As such, a WRC approval should not be necessary for this big bulk of code.

## Next steps

After this has been merged, I would like to extract `tnoodlejs` and `tnoodle-android` to a separate `tnoodle-examples` repository. That way, `thewca/tnoodle` really only consists of the core TNoodle scrambler for official WCA competitions and nobody has to install Android SDKs anymore.

That change includes migrating `ScrambleRequests.java` and breaking it down into more digestible bites. Please do not work on this now, the efforts would be redundant.

### Winstone
I am already working on a rewrite of the `webscrambles` module to use a more modern server core (Winstone has not been maintained since ~2009). That branch is already at a point where it's usable and clean, and everything server-related has been migrated to Kotlin. I plan to submit the PR as soon as this one is merged.

### TNoodle-LIB
When I was talking to Jeremy a few weeks ago we figured it would be nice to dissect the repository even further and extract the `scrambles` code into a separate repository so that we can change the webserver independently of how the scrambles are actually generated. A more detailed plan should be laid out in a separate issue.

## Closing words
I hope that this change makes it easier for a lot of people to contribute and work more on this piece of WCA Software. I have significant experience in Gradle setups, so if you run into trouble then don't hesitate to contact me. There is a public email attached to my GitHub account, and all the WCA Staff should know how to reach me anyways.

I will be travelling from June 6 to June 11, so that's the perfect opportunity for you to paste your comments and then afterwards I hope we can get this beauty merged upstream!

Bonus: This partially fixes #160 :D


campos20:
This is probably the most beautiful PR I have seen, thanks for your work.

We should do a checklist of features to be manually tested.

README.md
```diff
@@ -23,41 +23,54 @@ Officially, `TNoodle-WCA` is a [scramble program](https://www.worldcubeassociati
...
+
+    org.gradle.console=rich
 
 
 ### WCA Scramble Program
 
 When you're ready to develop, run the following and then visit <http://localhost:2014/scramble/>
 
-    ./tmt make run -p wca
+    ./gradlew :webscrambles:runShadow
```
jfly: Should this command work without setting up android? When I try it, I get the following output:

```
➜  ~/gitting/tnoodle git:(9254e38c) ./gradlew :webscrambles:runShadow

FAILURE: Build failed with an exception.

* Where:
Build file '/home/jeremy/gitting/tnoodle/build.gradle.kts' line: 31

* What went wrong:
A problem occurred configuring project ':tnoodle-android'.
> SDK location not found. Define location with an ANDROID_SDK_ROOT environment variable or by setting the sdk.dir path in your project's local properties file at '/home/jeremy/gitting/tnoodle/local.properties'.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 0s
```

I'd really like for it to be possible for developers to do non-android things and not have to install the android SDK.

build.gradle.kts
```diff
@@ -0,0 +1,50 @@
...
+
+val releasePrefix = "TNoodle-WCA"
+
+tasks.create<Copy>("generateOfficialRelease") {
+    description = "Generate an official WCA release artifact. THIS WILL RUN TESTS!"
+    group = "WCA"
+
+    val targetProject = "webscrambles"
+
+    dependsOn(getTasksByName("check", true))
```
jfly: Could we remove this? I don't think this is appropriate to do when building the asset. I'd prefer that we explicitly run `./gradlew check` and `./gradlew generateOfficialRelease` in Travis. I think this is also nice change to make for developers who just want to build a real jar file locally without waiting a very long time for tests to run, rather than forcing them to learn about yet another gradle task to run (`shadowJarOfficial`).

(don't forget to update this task description (it mentions tests) and the README as well)

buildSrc/src/main/kotlin/dependencies/Libraries.kt
```diff
@@ -0,0 +1,30 @@
+package dependencies
+
+object Libraries {
+    val GWTEXPORTER = "org.timepedia.exporter:gwtexporter:${Versions.GWTEXPORTER}"
```
jfly: Is the format of this Libraries.kt a standard Gradle thing? If not, do these constants get used anywhere?

.travis.yml
```diff
@@ -3,24 +3,11 @@ jdk: oraclejdk8
...
-  - nvm use v8.4
-  - bundle install
-  - (cd git-tools; npm install -d)
 script:
-  - ./tmt i18n check
-  - ./tmt lint -a
-  - ./tmt make dist -p wca
-  - ./tmt make check
-  - (cd tnoodle-ui; yarn test)
+  - ./gradlew publishToMavenLocal
```
jfly: Can we instead make `generateOfficialRelease` depend on `publishToMavenLocal`?

README.md
```diff
@@ -23,41 +23,54 @@ Officially, `TNoodle-WCA` is a [scramble program](https://www.worldcubeassociati
...
 
 ### Overview
 
-Get a high level view of all the projects that comprise tnoodle by running:
+Gradle is served through the use of a `Gradle wrapper` available as `gradlew` (UNIX systems) or `gradlew.bat` (that other operating system nobody wants to use)
+It is recommended to set up an alias to simplify task generation, along the lines of `alias gw='./gradlew --parallel'`.
+
+Get an overview of the core project tasks by executing
+
+    ./gradlew tasks
```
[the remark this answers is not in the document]
suushiemaniac: I opted for number 2.

README.md
```diff
@@ -23,41 +23,54 @@ Officially, `TNoodle-WCA` is a [scramble program](https://www.worldcubeassociati
...
 
-    ./tmt graph --descriptions
+If you want fancy console output, set the option `org.gradle.console` to value `rich`.
+This setting is currently not checked in because it makes Travis CI logs horrible to read, but you can easily include it by modifying (or creating, if it doesn't exist yet) `~/.gradle/gradle.properties` to something like
```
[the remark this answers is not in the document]
suushiemaniac: Huh, turns out this was a quirk with the terminal on my local machine.
Thanks for indirectly figuring that out for me :) I removed the corresponding README section.

README.md
suushiemaniac: That sounds nice indeed, but I'm afraid Gradle needs android to do some basic build configurations. There is no easy workaround that doesn't involve hijacking Gradle (which you don't want to, believe me).

All I can offer you as a "fix" is moving `tnoodlejs` and `tnoodle-android` to a separate repository (i.e. project) as soon as possible after merging this PR. That would also mean removing the Android plugin from the core Gradle build.

buildSrc/src/main/kotlin/dependencies/Libraries.kt
suushiemaniac: It is not *the* standard Gradle thing, because there is no common convention. It is however one of the formats commonly used in the real world to define a set of fixed version dependencies.

The constants get used all over the place in different `build.gradle.kts` scripts throughout the entire project.

build.gradle.kts
suushiemaniac: I have decoupled the task structure accordingly

.travis.yml
suushiemaniac: I especially like this idea because that way people don't really have to know anything about the artifact publication that is going on in the background.

README.md
jfly: SGTM!

jfly:
LGTM! Great work here. :shipit: :ship: IT

PR #420, opened 2019-07-29
Remove tnoodle-android and tnoodlejs

suushiemaniac:
They have both been moved to thewca/tnoodle-samples, including history and everything else.


suushiemaniac:
Please note, I will merge this up immediately as soon as the Travis build succeeds.

The code has been entirely migrated to thewca/tnoodle-samples so if anything goes wrong by even the slightest degree we can easily recover any lost code from there.

jfly:
Wooohooo!

PR #421, opened 2019-08-01
byebye-winstone

suushiemaniac:
Yay, here goes another big infrastructure commit!

# ktor
...is the name of an absolutely awesome server framework written in Kotlin. In contrast to `winstone`, it is actively maintained and much more flexible.

### Pros
+ We use a server framework that has active support via GitHub issues
+ Kotlin <3
+ TNoodle routing does not rely on XML merging any more
+ no more WEB-INF folders and nasty `tnoodle_resources` unpacking - we can use content directly from within the bundled JAR
+ no more silly `scrambler-interface` and `mootools` folders - everything is packaged into one static resource directory
+ no more superfluous `utils` and `web-utils` subprojects. The functions are all either redundant or merged into `webscrambles` now
 
### Cons
-- Kotlin stdlib (necessary for the program to execute at runtime) is a few MB worth of extra bytecode. I think it's fair to add it, since Kotlin will give us great benefits down the line
-- ktor was conceived to do much much more than the small bit of embedded server work we need it to. This means that a lot of additional dependencies are introduced into the tree
-- No "aggressive" server binding anymore, i.e. it's hard to launch two TNoodle instances in parallel

## side effects
- Migrated all `webscrambles` code to Kotlin :tada: 
- Made an effort to split up `ScrambleRequests`. In particular, PDFs are grouped and structured more finely now!
- BONUS: ProGuard :shield: static bytecode optimisation drops the JAR from 30MB (Kotlin+ktor deps included) down to ~15MB

## Important things to test
Please try out PDF generation like crazy. I did **not** touch any of the scrambling code so that part is fine, but image drawing and PDF rendering uses a hell lot of reflection that might break.
Please generate PDFs for all events and see if they look fancy. The actual scramble strings are not of interest for this PR and will be tested by Travis.


suushiemaniac:
@campos20 I had to manually migrate your "OrderedScrambles" fix from the 0.15 release (the one with `hasMultipleRooms || hasMultipleVenues || hasMultipleFooBar`) to port it over to Kotlin.

Please look at https://github.com/thewca/tnoodle/pull/421/commits/e13958f087f10fb8852554e7e0fdd62b8de4b0ea and check if it does what it's supposed to!

suushiemaniac:
Now that CI works again, I'd like to merge this up as soon as Travis says yes.

As with the other Gradle PR, the only real way to test this is to enable people to use it.
I would be willing to compile an unofficial release and take responsibility to have Delegates test whether it runs on their local machines.

scrambles/src/main/java/net/gnehzr/tnoodle/puzzle/CubePuzzle.java
```diff
@@ -50,7 +45,7 @@ public CubeMove(Face face, int dir, int innerSlice, int outerSlice) {
             this.innerSlice = innerSlice;
             this.outerSlice = outerSlice;
             // We haven't come up with names for moves where outerSlice != 0
-            azzert(outerSlice == 0);
+            assert outerSlice == 0;
```
jfly: Why the change here? IIRC, the reason we use `azzert` instead of `assert` is because we wanted to guarantee that assertions were enabled at runtime, and that wasn't possible to do because it's a parameter the person running TNoodle would need to tweak.

scrambles/src/main/java/net/gnehzr/tnoodle/puzzle/MegaminxPuzzle.java
```diff
@@ -83,7 +80,8 @@ public Dimension getPreferredSize() {
     private static final double UNFOLDWIDTH = 4 * Math.cos(.1 * Math.PI) + 2 * Math.cos(.3 * Math.PI);
 
     private static void turn(int[][] image, Face side, int dir) {
-        dir = GwtSafeUtils.modulo(dir, 5);
+        dir = ((dir % 5) + 5) % 5;
```
jfly: Just curious, why the change here?

scrambles/src/main/java/net/gnehzr/tnoodle/scrambles/Main.java
```diff
@@ -1,89 +0,0 @@
-package net.gnehzr.tnoodle.scrambles;
```
jfly: I'm ok with getting rid of this, but for the record, I think some people do use this to use TNoodle to generate scrambles in a headless way. We'll probably hear from the eventually. It would be nice for us to have a good answer for them when they do contact us =)

scrambles/src/test/java/net/gnehzr/tnoodle/puzzle/SquareOnePuzzleTest.java
```diff
@@ -41,6 +40,6 @@ public void testSomething() throws InvalidMoveException {
         solution = state.solveIn(2);
         assertEquals(solution, "(-3,1)");
 
-        System.out.println(sq1.generateWcaScramble(Utils.getSeededRandom()));
+        System.out.println(sq1.generateScramble());
```
jfly: Why the change here (and in 3x3 fewest moves test below)?

We went to a good bit of effort to make sure that tests were random, but reproducible so long as you use the same seed. It would be a shame to lose that.

scrambles/src/test/java/net/gnehzr/tnoodle/scrambles/AlgorithmBuilderTest.java
```diff
@@ -26,14 +23,14 @@ void testRedundantMoves() throws LazyInstantiatorException, InvalidMoveException
         assertFalse(moves.contains("3Lw"));
         assertFalse(moves.contains("3Dw"));
 
-        Map<String, LazyInstantiator<Puzzle>> lazyScramblers = PuzzlePlugins.getScramblers();
+        Map<String, Puzzle> lazyScramblers = new HashMap<>();
+        // FIXME Map<String, Puzzle> lazyScramblers = PuzzlePlugins.getScramblers();
 
-        for (Map.Entry<String, LazyInstantiator<Puzzle>> lazyEntry : lazyScramblers.entrySet()) {
+        for (Map.Entry<String, Puzzle> lazyEntry : lazyScramblers.entrySet()) {
```
jfly: This looks like a blocker: is this test testing anything now?

scrambles/src/main/java/net/gnehzr/tnoodle/puzzle/CubePuzzle.java
suushiemaniac: I was generally trying to get rid of GWTSafeUtils as part of the restructure.
Using a "forced" assert irrespective of the user settings never actually occured to me, but are you sure that enforcing the hard-wired fails is really necessary? In other words, does the assertion ever actually fail in production?

scrambles/src/main/java/net/gnehzr/tnoodle/puzzle/MegaminxPuzzle.java
suushiemaniac: Because I wanted to get rid of GwtSafeUtils, see above.

scrambles/src/main/java/net/gnehzr/tnoodle/scrambles/Main.java
suushiemaniac: The best answer will most likely be to use the webserver API in conjunction with any REST client that exists out there.
Possible answers are:
- `Insomnia` on Linux
- `Paw` on macOS
- `Don't use Windows(TM)` on Windows

scrambles/src/test/java/net/gnehzr/tnoodle/puzzle/SquareOnePuzzleTest.java
suushiemaniac: I was unaware of those efforts to be honest. We can revert this change at any given time if you feel it's necessary.

scrambles/src/main/java/net/gnehzr/tnoodle/puzzle/CubePuzzle.java
jfly: It's hard to know if they hit in production, because we'd only know if people tell us. I don't now of any recent situations in which that has happened. I'm sure it happened in the earlier days.

My thinking is: the code was written with these invariants in mind, and it's really hard to predict the behavior of the code if it keeps running past any of these assertions. It would be really bad an assertion failing to blow caused us to generate a scramble that doesn't obey the WCA regulations for some reason.

scrambles/src/test/java/net/gnehzr/tnoodle/scrambles/AlgorithmBuilderTest.java
suushiemaniac: Good catch! This `FIXME` really just slipped my intention because of the sheer size of this commit. I swear that this will be the last big infrastructure change for the forseeable future!!

suushiemaniac:
Wow, running ProGuard optimisations on TNoodle is actually killing Travis memory-wise. My two cents:
1. TNoodle has too many external dependencies
2. We can use a non-optimised JAR for the CI because the main intention is to see whether the code compiles. There are no integration tests (or even sanity checks!) being run on that output JAR, so the size (in terms of MB) doesn't matter here.

suushiemaniac:
Good point, guess I was too eager to get this thing merged already. I will pay more attention to future commit messages! :)

PR #422, opened 2019-08-14
WIP Test for quality of 3x3 scrambles

campos20:
This new project tests the quality of 3x3 scrambles.

In particular, I ran tests on the new min2phase and it passed.

# How it works

Run App.java and it will generate a lot of scrambles from [this function](https://github.com/thewca/tnoodle/blob/c6bd1ef69271932ddb285aa1c393c435c168b5cf/scrambles/src/main/java/net/gnehzr/tnoodle/scrambles/Puzzle.java#L99).

We tests scrambles for

- If the edges orientation distribution follows what is expected for random state
- If the corners orientation distribution follows what is expected for random state
- If the edges position distribution follows what is expected for random state
- If the corners position distribution follows what is expected for random state
- Cases with parity also follow that.

This project uses Chi Square Tests for edge and corner both orientation and position distribution. For parity, this uses Binomial Test. Both cases, we use alpha = 0.01, which means 99% precision.


suushiemaniac:
I have migrated the build project to Gradle over in https://github.com/campos20/tnoodle/pull/3

Full code review will be due once the big server infrastructure commit has been merged :)

suushiemaniac:
I believe this PR belongs to https://github.com/thewca/tnoodle-lib now.
Should I port it over?

campos20:
> I believe this PR belongs to https://github.com/thewca/tnoodle-lib now.
> Should I port it over?

I'll close this and do a copy / paste to keep you from manipulating trees.

suushiemaniac:
Actually, in this case it would be rather easy because all of the commits happen to be within a separate subfolder already

suushiemaniac:
Closing in favo(u)r of https://github.com/thewca/tnoodle-lib/pull/18

PR #426, opened 2019-08-15
Translate Registrant ID

campos20:


webscrambles/src/main/resources/tnoodle_resources/i18n/da.yml
```diff
@@ -20,6 +20,8 @@ da:
         result: Resultat
         #original_hash: 1917f4f
         competition: Konkurrence
+        #original_hash: ca817e3
+        registrantId: Registrerings-ID
```
DanielEgdal: Could we actually change this to "Tilmeldings-ID"? I realise this would be more consistent with how registration has been translated elsewhere on the website.

campos20: Fixed. Sorry for the delay

suushiemaniac:
@campos20 When do you expect to find time to work on this?

campos20:
> When do you expect to find time to work on this?

It looks like you hunt them down. I should complete this in the next 40 min (I'll make sure running is ok, that every translation fits what is expected and I'll squash the commits).

campos20:
I did not double checked yet. It will take a day or 2.

PR #430, opened 2019-10-10
Change 'large' number of multi cubes

jpbulman:
This was posted in a FB discussion

28 is arguably no longer a 'really large' number of cubes to attempt.

### DIFF
--- a/scrambler-interface/src/main/resources/tnoodle_resources/webapps/ROOT/scrambler-interface/js/ui.js
+++ b/scrambler-interface/src/main/resources/tnoodle_resources/webapps/ROOT/scrambler-interface/js/ui.js
@@ -11,7 +11,7 @@ var mark2 = {};
 (function() {
     "use strict";
 
-    var THREE_MULTI_DEFAULT_PUZZLES_PER_ATTEMPT = 28;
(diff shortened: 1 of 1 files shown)


campos20:
MBLD prints 7 scrambles per page so that's why we use 28. We should keep that. If we set this to 63 or 70, this could lead to waste of paper since 28 fits most competitors. Can you please consider this in your PR? Maybe just remove "really large" or update to a number > 28 but not as big as 100, multiple of 7.

suushiemaniac:
Ping @jpbulman

jpbulman:
Ah okay, I didn't notice the multiples of seven, that makes more sense now. I like the idea of replacing the message, maybe something along the lines of "Note: For multiple blindfolded, this is actually the number of scrambles you want generated for each attempt in the round! This should be set to the largest number of cubes a competitor will attempt within a round." ? I'd imagine keeping a default of 28 would probably be fine, that is still somewhat high and probably encompasses most MBLD competitors on average. Thoughts?

campos20:
On top of that, we'll soon replace the layout by a more uniform one (offline and online view). If it's ok for you, I'd like to close this.

jpbulman:
Sounds good!

PR #444, opened 2019-11-02
Add cloud capabilities to TNoodle-webscrambles

suushiemaniac:
# Introducing: GCloud

With this commit, TNoodle can be deployed to the [Google App Engine](https://cloud.google.com/appengine/)!

In particular, please refer to the [GCloud SDK docs](https://cloud.google.com/appengine/docs/standard/java/building-app/environment-setup) for deployment instructions. You will have to install your own `gcloud` sdk package.

A **proof of concept** can be seen at https://wca-scrambles-unofficial.appspot.com/ :tada: 

## Code changes

Mostly adding the Google Application Engine plugin to Gradle. There are some changes in the way `ktor` boots, which fortunately makes the code a little easier in some parts. We are now using a stand-alone resource configurattion file instead of Java code configurations :upside_down_face: 

The version information that was previously processed through the `MANIFEST.mf` file is now in a separate `tnoodle.version` resource file, because Google Cloud does not support manifests.

TNoodle server can now read blobs from Google Cloud Storage for pruning tables. Important resources like translations are still available from within the JAR

## Other notes

TNoodle can still be compiled into a standalone JAR that launches like usual. The Google Cloud functionality is **additional** to the existing structure

My first PR under 1000 lines of code :smile:


suushiemaniac:
Accidentally bloating up the release JAR with Google libraries is an interesting thought. I will contemplate some more on that issue.

@WST Please do not merge until I find a solution. If I should withdraw the PR at all, just leave a comment ;)

suushiemaniac:
@jfly I figured out a solution that I'd like your opinion on: Instead of managing everything through `webscrambles`, I have introduced a second subproject called `cloudscrambles` that manages the Google SDKs on top of `webscrambles`.

When building the JAR for Delegates, we run ProGuard over `webscrambles`, so the Google libraries are not included. `cloudscrambles` extends that basic module, and adds the necessary infrastructure for cloud deployments on top (in a second, independent subproject).

Different file access variants are managed through the interface `org/worldcubeassociation/tnoodle/server/util/ServerEnvironmentConfig.kt`. This is required as a constructor parameter of `TNoodleBaseServer` now, so that essentially whenever you want to spin up a TNoodle server you have to pass TNoodle an implementation of that interface so the server knows how to read/write files for example.

suushiemaniac:
Kindly ping @jfly, I'd like to get this thing merged before the end of the year :)

suushiemaniac:
Merging due to lack of additional feedback. We need this restructure for clean integration of `tnoodle-lib`.

PR #451, opened 2019-11-16
UpGradle version 6

suushiemaniac:
Literally just bumping a version string and then exploiting new Gradle 6 features to refactor copy-pasta code into a common helper class (now that it's possible to do so architecture-wise)

Beauty corrections of the build system, nothing major. Merging up as soon as Travis says Yes.

### DIFF
--- a/buildSrc/src/main/kotlin/configurations/CompilerSettings.kt
+++ b/buildSrc/src/main/kotlin/configurations/CompilerSettings.kt
@@ -1,0 +1,5 @@
+package configurations
+
+object CompilerSettings {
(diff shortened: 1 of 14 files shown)


suushiemaniac:
Heads-up: There is an issue in the `moowork` Gradle plugin that we're using to integrate `yarn` into the Java build. This currently blocks us from using Gradle 6 :(

suushiemaniac:
Heads-up: Some very nice people have forked the NPM plugin to provide Gradle 6 compatibility, because the owner of the original plugin seems pretty inactive.
Huge shout-out to https://github.com/node-gradle/gradle-node-plugin!

PR #453, opened 2019-11-17
Migrate Joda to J8 DateTime (fixes #400)

### LINKED ISSUE #400: Migrate from Joda to J8 Date/Time
Among other good new things java 8 has, we are using Joda-Time for dates, via its .jar file.
After we update to java8 and use java.time, which is the same class, as of JSR-310
We can dismiss joda-time-2.10.1.jar from /lib

suushiemaniac:
Addresses #400

We don't need Joda anymore now that Java has a reliable and (somewhat) clean DateTime implementation on board.

This also saves us one library that we would need to package for the Delegates otherwise.

### DIFF
--- a/buildSrc/src/main/kotlin/dependencies/Libraries.kt
+++ b/buildSrc/src/main/kotlin/dependencies/Libraries.kt
@@ -3,7 +3,6 @@ package dependencies
 object Libraries {
     val GWTEXPORTER = "org.timepedia.exporter:gwtexporter:${Versions.GWTEXPORTER}"
     val MARKDOWNJ_CORE = "org.markdownj:markdownj-core:${Versions.MARKDOWNJ_CORE}"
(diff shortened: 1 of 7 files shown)


campos20:
So far, the competitions I tried looks ok, but there's a thing. It was expected a single file with all scrambles ordered, but it's missing.

Proposed version

Captura de tela de 2019-12-07 01-29-42

Current official version

Captura de tela de 2019-12-07 01-37-28

suushiemaniac:
Thanks for catching that one, @campos20! I have re-tested and it seems that this is a general problem with the Kotlin transition, it has nothing to do with this commit.

You can see yourself by comparing the current tip of `master` to this PR. Both versions lack the "complete" file, so I kindly request you to file a separate issue for that and just accept this PR in the meantime.

campos20:
OK. I'll assign to that one. Seems a good opportunity to get started with kotlin

PR #456, opened 2019-12-08
Clearly distinguish clock pin positions in SVG

### LINKED ISSUE #423: Show raised clock pins in drawings
There have been recent clock pin issues, including some scrambles with completely inverted pins. It would be nice to improve the way we show pins that are up by extending them in the drawings:

Clock Round 1

suushiemaniac:
Fixes #423

Basically fakes the impression of a cylinder by stacking "bottom circle" + "body rectangle" + "top lid circle" on top of each other.

## FAQ
Q: Why not use "circle-rectangle-circle" alone? Why the additional path strokes?
A: See source code comments. Adding strokes to the central cylinder would also cross the bottom "cylinder foot" circle, and that would require yet another overlay circle.

Q: Where do you define the height of the pin?
A: Constant `int` at the top of the `ClockPuzzle` class, called `pinUpOffset`. The concrete value `6` just occured to me as a good approximation of Lucas' mockup in the original issue.

Q: Wait, you're saying clock pins are important?!
A: This PR is **not** a political statement in the 2020 Regulation Updates debate. I am merely trying to check off issues, any further implications would be purely coincidental.

### DIFF
--- a/scrambles/src/main/java/net/gnehzr/tnoodle/puzzle/ClockPuzzle.java
+++ b/scrambles/src/main/java/net/gnehzr/tnoodle/puzzle/ClockPuzzle.java
@@ -1,11 +1,7 @@
 package net.gnehzr.tnoodle.puzzle;
 
-import net.gnehzr.tnoodle.svglite.Color;
-import net.gnehzr.tnoodle.svglite.Dimension;
-import net.gnehzr.tnoodle.svglite.Circle;
-import net.gnehzr.tnoodle.svglite.Path;
-import net.gnehzr.tnoodle.svglite.Svg;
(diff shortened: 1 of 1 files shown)


suushiemaniac:
Here's a screenshot of the resulting SVG in action:
2019-12-08-125524_1531x751_scrot

suushiemaniac:
Here's a PDF with Clock-only scrambles:
[Scrambles for 2019-12-08 - All Scrambles.pdf](https://github.com/thewca/tnoodle/files/3936195/Scrambles.for.2019-12-08.-.All.Scrambles.pdf)

suushiemaniac:
Including this in the 2020 release just to get some feedback from competitions.

PR #458, opened 2019-12-15
Run npm update and npm audit fix

campos20:
This is a preparation for refactoring the UI.

### DIFF
--- a/tnoodle-ui/yarn.lock
+++ b/tnoodle-ui/yarn.lock
@@ -2,6 +2,13 @@
 # yarn lockfile v1
 
 
+"@babel/runtime@^7.1.2":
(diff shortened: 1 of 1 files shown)


campos20:
I'll replace by `yarn upgrade`. Now that you mentioned, it makes more sense since we already have a yarn.lock. Think should change about 2k lines and not create a 12k lines file.

campos20:
I replaced by `yarn upgrade`. I can't explain all the changes though, but it's a good thing to keep our dependencies updated. As far as I can tell, yarn does not have a `audit --fix` alternative.

suushiemaniac:
Dependency updates are fine by me!

PR #459, opened 2019-12-19
Move PuzzlePlugins to server package

suushiemaniac:
We use `PuzzlePlugins` to lazily provide puzzle instances that behave like singletons to the `webscrambles` server.

There is no reason why the `:scrambles` code (which is subject to be refactored into `tnoodle-lib`) should know these plugins. Also, changing this gets rid of the entire Kotlin dependency of the `:scrambles` module (for now, I'd rather migrate to Kotlin all at once).

The only reasonable alternative would be to move PuzzlePlugins back to Java entirely, but I currently don't see any reason why this would be necessary. Please advise.

### DIFF
--- a/scrambles/build.gradle.kts
+++ b/scrambles/build.gradle.kts
@@ -16,7 +16,6 @@ plugins {
     `java-library`
     checkstyle
     `maven-publish`
-    kotlin("jvm")
 }
 
 configureJava()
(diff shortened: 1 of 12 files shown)


suushiemaniac:
Merging in T-48 hours unless anyone objects.
This is currently keeping us from refactoring lib-tnoodle

PR #461, opened 2019-12-19
Store pruning tables for 4x4

suushiemaniac:
Turns out that due to a programming mistake, 4x4 pruning tables were never stored and therefore re-computed on every run of the program.

We can save considerable memory (especially in cloud environments) by storing the pruning files on disk.


campos20:
Interesting. 4x4 takes forever to generate. Can you also check sq-1 as well?

suushiemaniac:
This change _should_ actually stimulate the exact opposite behaviour. I'll keep an eye on it but it's not on top of my ToDo list right now :/

suushiemaniac:
Re-triggering…

suushiemaniac:
Labelling this as WIP since it's unclear whether the I/O of pruning tables actually provides an advantage over raw initialization every time

Issue #198, opened 2015-11-19
Password can be cracked easily


vojtechkral:
Hi.
I'm not sure if you're aware of this, but ZIP archive encryption is, well, shite. It suffers from the _known plaintext attack_.

__TL;DR:__ Any TNoodle-generated password-protected zip file can be cracked in a couple of minutes.

Longer version:
To crack a TNoodle zip file, all the attacker needs is to use TNoodle to create another zip file with the same competition name and use it for a know plaintext attack. The zip file the attacker creates must _not_ be password protected and it doesn't have to contain the same set of events as the zip file to be cracked. All that matters is that the competition name is same.

This creates and identical file in both the original encrypted zip file and the attacker's zip file. It's the html file (ie. `<competition name>.html`). This file can then be used to conduct the attack.

I've tried this on my machine using the `pkcrack` utility:

```
$ time ./pkcrack -C TramtariaOpen.zip -c TramtariaOpen.html -P myzip.zip -p TramtariaOpen.html -d cracked.zip
Files read. Starting stage 1 on Fri Nov 20 00:12:43 2015
Generating 1st generation of possible key2_298 values...done.
Found 4194304 possible key2-values.
Now we're trying to reduce these...
Done. Left with 28541 possible Values. bestOffset is 24.
Stage 1 completed. Starting stage 2 on Fri Nov 20 00:12:54 2015
Ta-daaaaa! key0= 101e707, key1=8da3970f, key2=60f27edf
Probabilistic test succeeded for 279 bytes.
Strange... had a false hit.
Strange... had a false hit.
Stage 2 completed. Starting zipdecrypt on Fri Nov 20 00:27:45 2015
Decrypting pdf/3x3x3 Cube Round 1.pdf (47baea80157c5a772e5050b1)... OK!
Decrypting txt/3x3x3 Cube Round 1.txt (57b856508580b941f3d750b1)... OK!
Decrypting TramtariaOpen.json (b619de8dd26ae8dccf7a50b1)... OK!
Decrypting TramtariaOpen.jsonp (56bdaaa55eaeb6c2632b50b1)... OK!
Decrypting TramtariaOpen.html (9bca3cddeaa5749400d450b1)... OK!
Decrypting TramtariaOpen.pdf (99253c5a6b0055e7ace750b1)... OK!
Finished on Fri Nov 20 00:27:45 2015
[...]

PR #463, opened 2019-12-19
Intentionally downgrade ZIP encryption algorithm

suushiemaniac:
With the newer version of the Java ZIP library we're using, we *technically* have the option to use strong AES encryption (cf. https://github.com/thewca/tnoodle/issues/198)

However, since the AES ZIP encryption algorithm was only retroactively added to the ZIP standard, many standard archiving libraries don't support it. Most notably, the native archive viewer on macOS as well as the `tar` CLI on Linux do **not** support AES.
You can always view the files (as the file headers in ZIP are unencrypted by definition) but you have to install third-party software to actually access their *content*.

Going forward, we have two options:
1. Use stronger AES encryption, thereby implicitly forcing Delegates to download and install third-party archiving tools such as 7zip that support extracting AES-encrypted ZIP files
2. Intentionally downgrade and accept the shortcomings of the weaker ZIP algorithm

Does anybody have favo(u)rable arguments for either option? I tend towards 2. as I'd like to keep the dependencies to run and use TNoodle at a bare minimum.

### DIFF
--- a/webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/ScrambleRequest.kt
+++ b/webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/ScrambleRequest.kt
@@ -168,7 +168,7 @@ data class ScrambleRequest(
 
             if (useEncryption) {
                 isEncryptFiles = true
-                encryptionMethod = EncryptionMethod.AES
+                encryptionMethod = EncryptionMethod.ZIP_STANDARD
             }
         }
 


suushiemaniac:
FYI @jfly @lgarron @campos20

EdHollingdale:
I favour 2 - Delegates use a range of operating systems and we should not force them towards anything for a marginal increase in security.

suushiemaniac:
In preparation for the 2020 release, I think it's best to include this.
If anyone has strong arguments in favo(u)r of AES encryption, then changing the encryption level is a one-liner for a potential 16.1 release

viroulep:
For the record: I think it's good we keep a broken encryption for our ZIP, although it's not intuitive.
In the current workflow Delegates use (or should use), all clear pdf are never given to scramblers at once, neither is the zip file.
Either they are printed, or the individual (properly) encrypted pdf are sent to the scramblers (or the laptop has its inputs locked and only the Delegate(s) can unlock and change the pdf).
Keeping broken encryption also allows us to recover the scrambles json when Delegates lost the password (which did happen several times in the past!).

PR #465, opened 2019-12-25
Bye-bye 333ft

suushiemaniac:
Removing 333ft from the TNoodle UI, because it is not recognised as an official event anymore starting January 1st, 2020.

There are no major implications from this change as far as I'm concerned, because the underlying 333 scrambler is still needed for plenty of other events.


suushiemaniac:
FYI: No idea why Travis is failing, presumably they cached some build files from my other "pruning" PR that should not be cached.

The changes are "harmless" and local builds do indeed proceed.

tnoodle-ui/src/wca/events.js
```diff
@@ -2,8 +2,8 @@
...
...Megaminx","format_ids":["a","3","2","1"],"can_change_time_limit":true,"is_timed_event":true,"is_fewest_moves":false,"is_multiple_blindfolded":false},{"id":"pyram","name":"Pyraminx","format_ids":["a","3","2","1"],"can_change_time_limit":true,"is_timed_event":true,"is_fewest_moves":false,"is_multiple_blindfolded":false},{"id":"skewb","name":"Skewb","format_ids":["a","3","2","1"],"can_change_time_limit":true,"is_timed_event":true,"is_fewest_moves":false,"is_multiple_blindfolded":false},{"id":"sq1","name":"Square-1","format_ids":["a","3","2","1"],"can_change_time_limit":true,"is_timed_event":true,"is_fewest_moves":false,"is_multiple_blindfolded":false},{"id":"444bf","name":"4x4x4 Blindfolded","format_ids":["3","2","1"],"can_change_time_limit":true,"is_timed_event":true,"is_fewest_moves":false,"is_multiple_blindfolded":false},{"id":"555bf","name":"5x5x5 Blindfolded","format_ids":["3","2","1"],"can_change_time_limit":true,"is_timed_event":true,"is_fewest_moves":false,"is_multiple_blindfolded":false},{"id":"333mbf","name":"3x3x3 Multi-Blind","format_ids":["3","2","1"],"can_change_time_limit":false,"is_timed_event":false,"is_fewest_moves":false,"is_multiple_blindfolded":true}].map(extend),
```
campos20: I don't think we should remove all the references foot 333ft here. From official, ok, but we should keep in the byId part. If you take a closer look, there are still magic, 333mbo...

suushiemaniac:
Merging up for the 2020 release

tnoodle-ui/src/wca/events.js
```diff
@@ -1,14 +1,361 @@
...
+    extend
+  )
 };
 
 function extend(rawEvent) {
   rawEvent = _.mapKeys(rawEvent, (v, k) => _.camelCase(k));
   return {
-    ...rawEvent,
-  }
+    ...rawEvent
```
viroulep: Was it removed automatically by your text editor?
In the website source we try to enforce the trailing commas in multilines arrays and map, it leads to cleaner diff when adding/removing elements.

PR #472, opened 2020-01-03
Adapt WCIF date parsing to escape from timezone hell

### LINKED ISSUE #471: Fix problem with timezones
Example

[AdelaideSummer2020](https://www.worldcubeassociation.org/competitions/AdelaideSummer2020#general-info) is a 1 day (January 11th) competition in Australia.

Captura de tela de 2020-01-01 18-19-31

It is expected that ordered scrambles respect that. It's timezone is Australia/Adelaide (UTC/GMT +10:30). From my point of view (GMT -3), it is a 2 days competition. There's a problem when generating Ordered Scrambles form a different timezone point of view.

From my point of view, 10:10 am is the last event in the day since it's actually [23:40 in GMT-3](https://www.timeanddate.com/worldclock/converter.html?iso=20200111T131000&p1=233&p2=5).

Captura de tela de 2020-01-01 18-24-12

Captura de tela de 2020-01-01 18-26-00

Day two starts with 11:10 event (therefore 00:40).

Captura de tela de 2020-01-01 18-27-28

We need to fix that behavior.

suushiemaniac:
Potentially fixes #471

### DIFF
--- a/webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/wcif/Activity.kt
+++ b/webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/wcif/Activity.kt
@@ -1,9 +1,8 @@
 package org.worldcubeassociation.tnoodle.server.webscrambles.wcif
 
-import java.time.LocalDateTime
 import java.time.ZoneId
-import java.time.ZonedDateTime
+import org.worldcubeassociation.tnoodle.server.webscrambles.wcif.WCIFHelper.Companion.parseWCIFDateWithTimezone
 
 data class Activity(val activityCode: String, val startTime: String) {
-    fun getLocalStartTime(timeZone: ZoneId) = ZonedDateTime.of(LocalDateTime.parse(startTime, WCIFHelper.WCIF_DATE_FORMAT), timeZone)
+    fun getLocalStartTime(timeZone: ZoneId) = startTime.parseWCIFDateWithTimezone(timeZone)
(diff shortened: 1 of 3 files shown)


campos20:
Look at the generation date before

Captura de tela de 2020-01-03 23-18-18

And now

Captura de tela de 2020-01-03 23-19-39

Also, LocalDate are cleaner to our case, since we just need yyyy-mm-dd, which is precisely what LocalDate is.

At least for AdelaideSummer your fix worked! I'm finishing the tests later though. Spending 5h to setup kotlin was not funny (issue with the kotlin eclipse plugin that made me download IntelliJ IDEA community; it's quite impressive how we can forget how to code by just switching IDEs).

suushiemaniac:
Oh boy. Thanks for your patience! What a bummer to hear that you struggled to set up Kotlin :disappointed: I have a good connection to some JetBrains guys, maybe I can convince them to give some more love to the Eclipse plugin.

Anyways, I've reverted back to LocalDate because your argument absolutely makes sense. The timestamp in the top-left corner is a completely unrelated issue, please disregard for this PR.

suushiemaniac:
FYI: PDF watermarking is fixed in #473

campos20:
I believe I'll start using kotlin on the native Idellij. Also, as a professor, I have access to the educational version (I use Eclipse just because I think I like it more). Anyways, I filled an issue there :)

Your code passed all scenarios I tested!

suushiemaniac:
Thank you for the review! (And for the attention of noticing this error in the first place)

PR #473, opened 2020-01-04
Use correct SimpleDateFormat for PDF watermarking

suushiemaniac:
As it turns out, uppercase `Y` is *week year* (very strange ISO calendar thing for finance world) whereas lowercase `y` is the Gregorian Calendar year that we're all used to.

Similarly, lowercase `m` is actually the *minute* whereas uppercase `M` should be used for *month*

### DIFF
--- a/server-ktor/src/main/kotlin/org/worldcubeassociation/tnoodle/server/util/WebServerUtils.kt
+++ b/server-ktor/src/main/kotlin/org/worldcubeassociation/tnoodle/server/util/WebServerUtils.kt
@@ -5,10 +5,11 @@ import java.net.URISyntaxException
 import java.nio.file.Files
 import java.nio.file.StandardCopyOption
 import java.text.SimpleDateFormat
+import java.time.format.DateTimeFormatter
 import java.util.Random
 
 object WebServerUtils {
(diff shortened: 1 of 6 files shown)


suushiemaniac:
Minuscule fix that can be merged as soon as Travis says Yes.
I'd reckon this change is on the same page as "translation updates" per our review policy.

suushiemaniac:
Very odd Travis fail, it looks like Maven Central was down for a few seconds. Compiling Kotlin works perfectly fine, triggering another build now

suushiemaniac:
Apparently, Maven Central is still broken :O
Waiting for a bit before I re-trigger

campos20:
What about that LocalDate and dismiss that SDF for now?

suushiemaniac:
Well, we need some kind of "Formatter" for printing Java date representations into human-readable form -- regardless of whether we use `java.util.Date` or `java.time.LocalDate`.

I went ahead and replaced the former implementation with the latter.

campos20:
LocalDate.toString should suffice

suushiemaniac:
Merging based on @campos20's positive review and the fact that I addressed all of his suggested changes.

Issue #373, opened 2018-10-19
Scrambling scape param


jdediegomrt:
Working with tnoodle android library I have found an issue. 
In android library, for Android _cancel(**boolean** mayInterruptIfRunning)_ doesn't necessarily stop the execution of the scrambling process. The AsyncTask only executes onCancelled(), and won't run onPostExecute() when it completes.

for this reason if you try to cancel an scramble it wont be cancelled. 

This could be solved adding a scape flag or somenthing similar into scrambling loops. Something like:

In Puzzle.java

```
package net.gnehzr.tnoodle.scrambles;

//This variable is my example
   private boolean skipPuzzleScramble = false;
   /*Add get & set*/

```

...

```
    public final String generateWcaScramble(Random r) {
        PuzzleStateAndGenerator psag;
        do {
            psag = generateRandomMoves(r);
        } while(psag.state.solveIn(wcaMinScrambleDistance - 1) != null && !skipPuzzleScramble);
        return psag.generator;
    }
```

...

```
    @NoExport
    public PuzzleStateAndGenerator generateRandomMoves(Random r) {
        AlgorithmBuilder ab = new AlgorithmBuilder(this, MergingMode.NO_MERGING);

        while(ab.getTotalCost() < this.getRandomMoveCount()) {
            HashMap successors = ab.getState().getScrambleSuccessors();

            try {
                String move;
                do {
                    move = (String)GwtSafeUtils.choose(r, successors.keySet());
                    successors.remove(move);
                } while(ab.isRedundant(move) && !skipPuzzleScramble );

                ab.appendMove(move);
            } catch (InvalidMoveException var6) {
                l.log(Level.SEVERE, "", var6);
                GwtSafeUtils.azzert(false, var6);
                return null;
            }
        }

        return ab.getStateAndGenerator();
    }

```

A simple way to check this issue is:
in tnoodle android project launched,
scramble a Square-1 or a 4x4x4 and quickly scramble a 2x2x2 or a Skewb
You will see that it takes too much time to end a scramble that initially takes a few second to end.

do you see this issue?
is this a correct way to make it possible?

jfly:
Ah, interesting! +@plusCubed to see if he has thought about this at all.

In general, there are a lot of places throughout TNoodle where we sit in a loop, CPU bound. It seems like a big pain to try to add "escape hatches" to each of them, and furthermore, to remember to add them to new places.

suushiemaniac:
[...]

PR #474, opened 2020-01-04
Turn ScrambleCacher work into Daemon threads

suushiemaniac:
One day I was debugging PDFs by using a simple custom main method:
1. Create dummy `ScrambleRequest`
2. Compile PDF, write to temporary debug file
3. Shut down

...except that the JVM was not shutting down because the ScrambleCacher threads were still running in the background. This is no big deal for the Delegates' JAR, because they would manually kill the server anyways, but it is annoying when using `scrambles` (prospective `lib-tnoodle`) as a standalone module.

If we want people to be able to use `ScrambleCacher` without much hassle (see for example #373) then we _should_ mark it as Daemon so the JVM can shut down gracefully if no other thread accesses the `ScrambleCacher` any more.

### DIFF
--- a/scrambles/src/main/java/net/gnehzr/tnoodle/scrambles/ScrambleCacher.java
+++ b/scrambles/src/main/java/net/gnehzr/tnoodle/scrambles/ScrambleCacher.java
@@ -87,6 +87,7 @@ public class ScrambleCacher {
                 }
             }
         });
+        t.setDaemon(true);
         running = true;
         t.start();
     }


suushiemaniac:
@jfly `git blame` tells me that you did virtually all of the current `ScrambleCacher` implementation. What's your take on this? Have you originally considered daemon threads when writing the first implementation?

jfly:
LGTM! This sounds very reasonable to me, good call!

I'm sure I did not consider this when building TNoodle. I always wanted to use TNoodle as a library, but only ever *actually* used it as a library with the GWT generated javascript, which would not have had this problem (because that generated code doesn't have any sort of threading/scramble cacher mechanism at all).

PR #476, opened 2020-01-08
Move pruning folder code to separate feature branch

suushiemaniac:
Right now, we create an empty folder that is unused, because pruning is only (subject to being) added in #461 

Until the .15 releases, this folder used to be necessary for resources. But now, we can access these directly from within the JAR so we should not create an empty folder.

Merging as soon as Travis says Yes because this is only moving "dead code" to another branch that will then be reviewed again more thoroughly.


suushiemaniac:
Retry after merging #477

suushiemaniac:
Okay, only reasonable fix seems to be brute force retry for now :/

PR #487, opened 2020-01-18
Refactor FMC moves table and fit text methods

campos20:
The main idea here is to render the moves table in a similar way we do to the other texts (positioning on rects). This will make it easier to maintain.

### DIFF
--- a/webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/FmcScrambleCutoutSheet.kt
+++ b/webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/FmcScrambleCutoutSheet.kt
@@ -1,9 +1,6 @@
 package org.worldcubeassociation.tnoodle.server.webscrambles.pdf
 
-import com.itextpdf.text.Document
-import com.itextpdf.text.Element
-import com.itextpdf.text.Image
-import com.itextpdf.text.Rectangle
+import com.itextpdf.text.*
 import com.itextpdf.text.pdf.PdfWriter
 import org.worldcubeassociation.tnoodle.server.webscrambles.ScrambleRequest
 import org.worldcubeassociation.tnoodle.server.webscrambles.Translate
@@ -12,7 +9,7 @@ import org.worldcubeassociation.tnoodle.server.webscrambles.pdf.util.PdfDrawUtil
 import org.worldcubeassociation.tnoodle.server.webscrambles.pdf.util.PdfDrawUtil.populateRect
 import org.worldcubeassociation.tnoodle.server.webscrambles.pdf.util.FontUtil
 
-class FmcScrambleCutoutSheet(request: ScrambleRequest, globalTitle: String?): FmcSheet(request, globalTitle) {
+class FmcScrambleCutoutSheet(request: ScrambleRequest, globalTitle: String?) : FmcSheet(request, globalTitle) {
     override fun PdfWriter.writeContents(document: Document) {
         for (i in scrambleRequest.scrambles.indices) {
             addFmcScrambleCutoutSheet(document, scrambleRequest, title, i)
@@ -51,9 +48,11 @@ class FmcScrambleCutoutSheet(request: ScrambleRequest, globalTitle: String?): Fm
 
         val paddedTitleItems = textList.zip(alignList)
 
+        val font = Font(BASE_FONT, FONT_SIZE)
+
         for (i in 0 until SCRAMBLES_PER_SHEET) {
             val rect = Rectangle(LEFT.toFloat(), (top - i * availableScrambleHeight).toFloat(), (right - dim.width - SPACE_SCRAMBLE_IMAGE).toFloat(), (top - (i + 1) * availableScrambleHeight).toFloat())
-            directContent.populateRect(rect, paddedTitleItems, BASE_FONT, FONT_SIZE)
+            directContent.populateRect(rect, paddedTitleItems, font)
 
             directContent.addImage(Image.getInstance(tp), dim.width.toDouble(), 0.0, 0.0, dim.height.toDouble(), (right - dim.width).toDouble(), top.toDouble() - (i + 1) * availableScrambleHeight + (availableScrambleHeight - dim.getHeight()) / 2)
 
@@ -72,7 +71,7 @@ class FmcScrambleCutoutSheet(request: ScrambleRequest, globalTitle: String?): Fm
         val SPACE_SCRAMBLE_IMAGE = 5 // scramble image won't touch the scramble
         val SCRAMBLE_IMAGE_PADDING = 8 // scramble image won't touch the dashed lines
 
-        val FONT_SIZE = 20
+        val FONT_SIZE = 20f
 
         val SCRAMBLES_PER_SHEET = 8
     }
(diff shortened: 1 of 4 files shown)


campos20:
This should be a 2 commits PR. I'm not sure why the first three are showing as new commits since they are already on master.

campos20:
Pinging @suushiemaniac for a kt review

suushiemaniac:
For the mysterious additional commits, switch to master and do:
`git fetch upstream` (or replace `upstream` by whatever your master TNoodle repo is)
`git rebase upstream/master`
`git switch feature/refactor-fmc`
`git rebase master`

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/FmcScrambleCutoutSheet.kt
```diff
@@ -71,16 +72,15 @@ class FmcScrambleCutoutSheet(request: ScrambleRequest, globalTitle: String?): Fm
     }
 
     companion object {
-        val BASE_FONT = FontUtil.getFontForLocale(Translate.DEFAULT_LOCALE)
+        val baseFont = FontUtil.getFontForLocale(Translate.DEFAULT_LOCALE)
+        val font = Font(baseFont, 20f)
```
suushiemaniac: 1. constant values in (companion) objects are written in SNAKE_CASE by convention, because they effectively act as constants.
2. `font` objects are bound to their particular PDF in itext5. I don't see why creating a font out of the baseFont object here is beneficial...

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/FmcSolutionSheet.kt
```diff
@@ -239,118 +250,109 @@ open class FmcSolutionSheet(request: ScrambleRequest, globalTitle: String?, loca
...
             }
+            rotationsList.add(list)
         }
+        movesCell.add(rotationsList)
+
+        // Table
+        val tableWidth = competitorInfoLeft - left - 2 * margin
+        val tableHeight = movesRect.height
+        val tableLines = movesType.size * (directionModifiers.size + 1) // F, F2, F' (direction modifiers) and the title
+        val cellWidth = 25f // TODO calculate
```
suushiemaniac: Are you leaving this TODO in on purpose?

campos20:
> For the mysterious additional commits, switch to master and do:
> `git fetch upstream` (or replace `upstream` by whatever your master TNoodle repo is)
> `git rebase upstream/master`
> `git switch feature/refactor-fmc`
> `git rebase master`

Thanks! I thought a

> git pull upstream master

then start a new branch from there would suffice.

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/FmcSolutionSheet.kt
```diff
@@ -366,16 +368,21 @@ open class FmcSolutionSheet(request: ScrambleRequest, globalTitle: String?, loca
...
+        val leadingMultiplier = 1.5f
+        val rulesFont = Font(bf, 15f)
         val rules = rulesList.joinToString("\n")
-        cb.fitAndShowText(rules, bf, rulesRectangle, 15f, Element.ALIGN_JUSTIFIED, leadingMultiplier)
+        cb.fitAndShowText(rules, rulesRectangle, rulesFont, Element.ALIGN_JUSTIFIED, leadingMultiplier)
 
         doc.newPage()
     }
 
+    private fun debugRect(cb: PdfContentByte, rectangle: Rectangle) {
```
suushiemaniac: Are you sure leaving in a debug method is a good idea?

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/FmcSolutionSheet.kt
```diff
@@ -239,118 +250,109 @@ open class FmcSolutionSheet(request: ScrambleRequest, globalTitle: String?, loca
...
+        val movesTypeFont = Font(bf, firstColumnFontSize, Font.BOLD)
 
-            var cell = PdfPCell(Phrase(movesType[i], Font(bf, firstColumnFontSize, Font.BOLD)))
-            cell.fixedHeight = cellHeight.toFloat()
-            cell.verticalAlignment = Element.ALIGN_MIDDLE
-            cell.horizontalAlignment = Element.ALIGN_RIGHT
-            cell.border = Rectangle.NO_BORDER
-            table.addCell(cell)
+        // Render moves
+        for (i in movesType.indices) {
```
suushiemaniac: As far as I can tell, you are completely dropping the table mechanism in favor of hand-painted and pre-calculated rectangles.

This is not a good idea in my opinion. First, using a table makes sense when the end result should indeed look like a table. Second, it makes the code much harder to maintain since most of the drawing will likely break if you just change one small number somewhere.

suushiemaniac:
> Thanks! I thought a
> 
> > git pull upstream master
> 
> then start a new branch from there would suffice.

The command you mention only brings in the upstream changes to its own branch. You have to explicitly put your branch on top of that, that's what all the rebasing is for

suushiemaniac:
I have a PR for PDF refactoring ready as well, I'd prefer to merge that in (20+ commits) and then have you rebase your ideas on top of that

suushiemaniac:
I'll look into it right after breakfast, for now please hold back these changes...

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/FmcSolutionSheet.kt
campos20: This is actually the main ideia. I tried to generate a sample including some moves and I found that do that using the table cells ended up with empty results. My test consisted in adding a new set of moves, namely, wide moves.

suushiemaniac:
Now I've uploaded all of my local code changes, you can go ahead and rebase.
Sorry for the resulting conflict, I guess I should update this repository with my work more often…

suushiemaniac:
I also totally get your idea about avoiding empty cells. But you should be able to cover those by using the `colspan` attribute like for example in [L302](https://github.com/thewca/tnoodle/blob/master/webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/FmcSolutionSheet.kt#L302)

campos20:
The changes so far have no visible effects

Captura de tela de 2020-01-18 23-01-15

Your refactoring fixed the problem I was having of blank table while adding moves

campos20:
This PR became more about passing a font instead of base font + size and calculating the width for the moves table. Also, @suushiemaniac, any particular reason why ou moved Solution Sheet outside FMC dedicated folder? I believe this was a mistake and I moved it back.

suushiemaniac:
Yes, I suppose this was an honest mistake. Thanks for noticing!

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/FmcSolutionSheet.kt
```diff
@@ -266,9 +256,25 @@ open class FmcSolutionSheet(request: ScrambleRequest, globalTitle: String?, loca
...
+        val tableLines = movesType.size * (direction.size + 1) // +1 for the title
+        val firstColumnWidth = tableWidth / 2
+        val cellWidth = (tableWidth - firstColumnWidth) / WCA_MOVES.size
+        val cellHeight = tableHeight / tableLines
+        val columns = WCA_MOVES.size + 1 // +1 for the first column
+        val movesFontSize = 10
+        val movesFont = Font(bf, movesFontSize.toFloat())
+
+        val table = PdfPTable(columns).apply {
+            setTotalWidth(floatArrayOf(firstColumnWidth.toFloat(), *FloatArray(WCA_MOVES.size) { it -> cellWidth.toFloat() }))
```
suushiemaniac: First time I'm seeing a practical use for *array unpacking! :D

PR #489, opened 2020-01-18
Clean up PDF generation methods

suushiemaniac:
Two major goals:
1. Get rid of `var`, because stateful programming is dangerous
2. Get rid of `tableAndHighlighting`, because it can be effectively pre-computed


suushiemaniac:
Travis choking again…

PR #492, opened 2020-01-18
Fix minor PDF quirks after refactor

suushiemaniac:
Found some small (but amusing...) issues while testing, these should be corrected now.


suushiemaniac:
Travis Maven choke

PR #498, opened 2020-01-22
Say Hello to TNoodle-lib

suushiemaniac:
With this commit. the scrambling logic and the webserver code will finally be separated.

The only line that glues them together (literally, it's just this one single line!!) is `implementation("org.worldcubeassociation.tnoodle:lib-scrambles:0.16.3")` in the `dependencies {}` block of `build.gradle.kts` under the `webscrambles` module

The builds are hosted at our newly-created Bintray account: https://bintray.com/thewca/tnoodle-lib/scrambles and they will eventually be mirrored to JCenter or even Maven Central


buildSrc/src/main/kotlin/configurations/Languages.kt
```diff
@@ -18,22 +18,12 @@ object Languages {
     fun Project.attachRemoteRepositories() {
         repositories {
             mavenCentral()
+            maven(url = "https://dl.bintray.com/thewca/tnoodle-lib")
```
suushiemaniac: Will eventually become "fully" public, i.e. mirrored to the central repositories that don't require a custom link to our specific Bintray.

Setting up these mirrors has been initiated, but it takes a lot of processing time (up to 48 hours initially) because of how Maven works.

_Technically_, the package already is public, you just need to specifically point your build tool of choice to this specific (but public!) URL

build.gradle.kts
```diff
@@ -119,11 +118,9 @@ tasks.create("generateDebugRelease") {
 }
 
 tasks.create("deployToCloud") {
-    dependsOn(getTasksByName("publishToMavenLocal", true))
     dependsOn("registerCloudReleaseTag", ":cloudscrambles:appengineDeploy")
```
viroulep: Not fully related to that PR, but I'm very curious to see all these cloud-related tasks.
What's the idea behind that and the concrete use-case?
I'm not sure I'm a huge fan of having google dependencies for no apparent reasons :s
Happy to hear more about that!

buildSrc/src/main/kotlin/configurations/Languages.kt
viroulep: Very naive question: in case bintray goes down, is there a way to make it point to a local copy of tnoodle-lib?

buildSrc/src/main/kotlin/dependencies/Libraries.kt
```diff
@@ -24,6 +23,7 @@ object Libraries {
     val WCA_I18N = "com.github.thewca:wca_i18n:${Versions.WCA_I18N}"
     val GOOGLE_APPENGINE_GRADLE = "com.google.cloud.tools:appengine-gradle-plugin:${Versions.GOOGLE_APPENGINE_GRADLE}"
     val GOOGLE_CLOUD_STORAGE = "com.google.cloud:google-cloud-storage:${Versions.GOOGLE_CLOUD_STORAGE}"
+    val TNOODLE_SCRAMBLES = "org.worldcubeassociation.tnoodle:lib-scrambles:${Versions.TNOODLE_SCRAMBLES}"
```
viroulep: Another very naive question: is there a way to analyse our dependencies and see which one are used/unused?
(maybe you did run something like that to remove gwtexporter :))

viroulep:
Oh and another related comment: do we have open PR here that addresses issues in TNoodle-lib?
The question also applies to issues related to TNoodle-lib, which we should probably migrate over there (it's possible to do so manually on each issue's page, but I'm not aware of something automatic for a bulk set of issues matching a label for instance).

buildSrc/src/main/kotlin/configurations/Languages.kt
suushiemaniac: Yes, that is easily possible by adding a local `path/to/repo` replication here. However, I highly doubt Bintray will go down for > 30 seconds at once that easily

build.gradle.kts
suushiemaniac: This was a proof-of-concept thing I once did in October/November to procrastinate some uni work. Right now, this is not official and the Google deps are **not** included in the JAR that we distribute to Delegates.

suushiemaniac: Check out https://wca-scrambles-unofficial.appspot.com/ for the cloud version in action

buildSrc/src/main/kotlin/dependencies/Libraries.kt
suushiemaniac: Well, since all dependencies are refactored into variables here, you can just use your favorite IDE's "show unused variables" inspection.
Please note, however, that with the way Gradle works these constants are just *links to the dependency*. Having an "unused" dependency in Gradle is as unharmful as pasting the URL `https://worldcubeassociation.org` in some arbitrary document.

suushiemaniac:
Thanks for the review and credit, @viroulep :) I've addressed your concerns in the individual commits, feel free to catch up either here or on Slack.

Good point about migrating issues, I'll look into that after my exams are over in 10 or so days

PR #499, opened 2020-01-25
Make scramble sheet ratio aware

campos20:
Everything should look almost like the same.


campos20:
[Testação groß teste à teste 2020 - All Scrambles.pdf](https://github.com/thewca/tnoodle/files/4111579/Testacao.gross.teste.a.teste.2020.-.All.Scrambles.pdf)

campos20:
Super long names are a problem.

suushiemaniac:
Is the testação PDF from *before* or *after* applying your changes?

campos20:
> Is the testação PDF from _before_ or _after_ applying your changes?

From after

suushiemaniac:
In that case, I'm afraid I have to point out several problems with this PR:
1. The page margins at the left and right side seem **extremely** narrow. Most modern printers can't even print this close to the border of the (physical) page

suushiemaniac:
I took the freedom to merge this as part of my cleanup session :D

PR #502, opened 2020-02-07
Strip diacritics from ZIP filenames

suushiemaniac:
This is the bottom line of a lengthy discussion about ZIP UTF-8 support, cf. https://github.com/srikanth-lingala/zip4j/issues/131


campos20:
Maybe it's worth checking

https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#stripAccents-java.lang.String-

suushiemaniac:
Nice, I didn't know that one. Thanks @campos20 :)

PR #510, opened 2020-02-11
Use new tnoodle-lib release to redirect scrambler output

suushiemaniac:


campos20:
That's great. Do you have an use in mind for that file?

suushiemaniac:
Right now, it seems that the only practical purpose is to reduce cluttering `STDOUT`.
However, logback allows filtering pretty much any log message to a std file so we may use this for debugging information in the future. :)

PR #528, opened 2020-02-20
Add potential support for staging watermark in PDF

suushiemaniac:
This feature is **NOT** ported to the old frontends. Effectively, it is not in use right now, but the new `awesome-ui` may incorporate it later.


suushiemaniac:
This is how it would look like… 
2020-02-20-202156_1920x1080_scrot
[Scrambles for 2020-02-20 - All Scrambles.pdf](https://github.com/thewca/tnoodle/files/4232599/Scrambles.for.2020-02-20.-.All.Scrambles.pdf)

PR #539, opened 2020-02-28
Reformat WCIFScrambleMatcher

suushiemaniac:
Attempt to make the core matching code and indexing behaviour more readable


server-ktor/src/main/resources/rsa/tnoodle_private.pkcs8.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
viroulep, line 1: @suushiemaniac please tell me you regenerated a new key :p

suushiemaniac:
@viroulep Yes absolutely! The updated gitignore was tracked on another branch, so when I developed multiple features at once this key slipped through while switching between them.

I had previously already added a "this is not dangerous" note to the other PR. For one, these are testing keys anyways and for two, I have generated a new one.

PR #541, opened 2020-02-29
Add Persons to WCIF model for better FMC language detection

suushiemaniac:
Only render FMC translations for languages from countries that have at least one registered competitor


viroulep:
Could this possibly be exposed as an option?

> [ ] only generate FMC translation for ....

With the text containing the actual list of detected translations, so that the use is clearly aware of what's going on.

suushiemaniac:
Merging as a silent API endpoint. No behaviour is changed by default unless you explicitly opt to use the data API

PR #543, opened 2020-03-02
Fix errors related to MBLD loading

suushiemaniac:
The UI fix is more general, but it was observed/encountered with Multi because Multi is one of the few events where the event with the most # of attempts is not commonly used at comps.


campos20:
Nice catch. Thanks for fixing!

PR #500, opened 2020-01-27
Pre calculate fontSize for all scrambles fix #497

campos20:
This a quick fix and is not part of the big refactoring I'm doing #499

@Laura-O 's scrambles should look like this

[Scrambles for Berlin Winter Cubing 2020 - All Scrambles.pdf](https://github.com/thewca/tnoodle/files/4142047/Scrambles.for.Berlin.Winter.Cubing.2020.-.All.Scrambles.pdf)

A complete set of scrambles should look like this

[Scrambles for 2020-01-31 - All Scrambles.pdf](https://github.com/thewca/tnoodle/files/4142030/Scrambles.for.2020-01-31.-.All.Scrambles.pdf)

The fix idea is pretty simple: pre calculate font size using scrambleRequest.allScrambles instead of each part. I also set minimum one line font size to 15 because 12 looked unnecessarily tiny.

PR #557, opened 2020-03-23
Remove some PDF's hack, simplify some parts of the PDF generation

campos20:
Updated pdf (as of 2020-04-21)

[Scrambles for 2020-04-22 - All Scrambles.pdf](https://github.com/thewca/tnoodle/files/4513530/Scrambles.for.2020-04-22.-.All.Scrambles.pdf)


campos20:
Apparently, sq1 hasn't changed. I'm not sure why I thought otherwise

https://github.com/thewca/tnoodle/pull/500#issue-367281335

campos20:
Perhaps we should print square 1 side by side though.  Small scrambles allow good image.

suushiemaniac:
What do you mean by "side by side"?

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/GeneralScrambleSheet.kt
```diff
@@ -33,32 +32,27 @@ class GeneralScrambleSheet(scrambleSet: ScrambleSet, activityCode: ActivityCode)
...
 
         // We don't let scramble images take up too much of a the page
-        val maxScrambleImageWidth = (availableWidth / 2).toInt().takeUnless {
-            // TODO - If we allow the megaminx image to be too wide, the
-            //   Megaminx scrambles get really tiny. This tweak allocates
-            //   a more optimal amount of space to the scrambles. This is possible
-            //   because the scrambles are so uniformly sized.
-            scramblingPuzzle.shortName == "minx"
-        } ?: 190
+        val maxScrambleImageWidth = (availableWidth / 3).toInt()
```
suushiemaniac: Where did you take that "a third of available width" from (as opposed to the previous "half of"), and how does it relate to the 190 constant for Megaminx?

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/GeneralScrambleSheet.kt
```diff
@@ -80,43 +74,20 @@ class GeneralScrambleSheet(scrambleSet: ScrambleSet, activityCode: ActivityCode)
...
-
-        val availableArea = Rectangle(scrambleColumnWidth - 2 * SCRAMBLE_PADDING_HORIZONTAL,
-            (availableScrambleHeight - SCRAMBLE_PADDING_VERTICAL_TOP - SCRAMBLE_PADDING_VERTICAL_BOTTOM))
-
-        val longestScramble = scrambles.maxBy { it.length } ?: ""
-
-        val longestAlignedScramble = scrambles.map { StringUtil.padTurnsUniformly(it, WIDEST_CHAR_STRING) }
-            .maxBy { it.length } ?: longestScramble
+    private fun getFontConfiguration(availableArea: Rectangle, scrambles: List<String>): Font {
+        val longestScramble = scrambles.flatMap { it.split(NEW_LINE) }.maxBy { it.length } ?: ""
```
suushiemaniac: There is `String?.orEmpty()` in Kotlin. So this can be `maxBy { it.length }.orEmpty()`

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/GeneralScrambleSheet.kt
```diff
@@ -80,43 +74,20 @@ class GeneralScrambleSheet(scrambleSet: ScrambleSet, activityCode: ActivityCode)
...
-        val fontSizeForMaskedUnaligned = PdfUtil.fitText(Font(FontUtil.MONO_FONT), longestScrambleMasked, availableArea, FontUtil.MAX_SCRAMBLE_FONT_SIZE, false, 1f) // FIXME const
-
-        // If the scramble contains newlines, then we *only* allow wrapping at the newlines.
-        val longestRespectingNewlines = longestAlignedScrambleMasked.takeIf { longestScrambleOneLine }
-            ?: longestAlignedScrambleMaskedAndStuffed
-        val fontSizeIfIncludingNewlines = PdfUtil.fitText(Font(FontUtil.MONO_FONT), longestRespectingNewlines, availableArea, FontUtil.MAX_SCRAMBLE_FONT_SIZE, true, 1f) // FIXME const
-
-        val oneLine = longestScrambleOneLine && fontSizeForMaskedUnaligned >= FontUtil.MINIMUM_ONE_LINE_FONT_SIZE
-        val perfectFontSize = fontSizeForMaskedUnaligned.takeIf { oneLine } ?: fontSizeIfIncludingNewlines
+        val perfectFontSize = min(fontSize, fontSizeIfIncludingNewlines)
```
suushiemaniac: You are basically discarding the distincion between one-line scrambles and line-wrapped scrambles. Can you add w few words on why you think this is okay?

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/GeneralScrambleSheet.kt
```diff
@@ -33,32 +32,27 @@ class GeneralScrambleSheet(scrambleSet: ScrambleSet, activityCode: ActivityCode)
...
 
         val scrambleImageSize = scramblingPuzzle.getPreferredSize(maxScrambleImageWidth, maxScrambleImageHeight)
 
+        val allScrambleStrings = scrambleSet.allScrambles.toPDFStrings(scramblingPuzzle.shortName)
+
         val scrambleImageHeight = scrambleImageSize.height.toFloat()
         val scrambleColumnWidth = availableWidth - indexColumnWidth - scrambleImageSize.width
 
-        val allScrambleStrings = scrambleSet.allScrambles.toPDFStrings(scramblingPuzzle.shortName)
+        val availableArea = Rectangle(scrambleColumnWidth, scrambleImageHeight - 2 * SCRAMBLE_MARGIN)
```
suushiemaniac: I see that you are subtracting the vertical margin here that was previously subtracted within the `getFontConfiguration` method.

If you look into the previous (before your changes in this PR) version of `getFontConfiguration`, you will see there are a lot of other margins and paddings being thrown around as well (`SCRAMBLE_PADDING_VERTICAL_TOP`, `SCRAMBLE_PADDING_VERTICAL_BOTTOM`, and others). What happened to them?

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/GeneralScrambleSheet.kt
```diff
@@ -80,43 +74,20 @@ class GeneralScrambleSheet(scrambleSet: ScrambleSet, activityCode: ActivityCode)
...
-        val perfectFontSize = fontSizeForMaskedUnaligned.takeIf { oneLine } ?: fontSizeIfIncludingNewlines
+        val perfectFontSize = min(fontSize, fontSizeIfIncludingNewlines)
 
         return Font(FontUtil.MONO_FONT, perfectFontSize, Font.NORMAL)
     }
 
     private fun requiresHighlighting(scrambleColumnWidth: Float, scrambleFont: Font, scrambles: List<String>): Boolean {
-        val paddedScrambles = scrambles.map { StringUtil.padTurnsUniformly(it, PdfUtil.NON_BREAKING_SPACE.toString()) }
-        val lineChunks = paddedScrambles.map { it.splitToLineChunks(scrambleFont, scrambleColumnWidth) }
+        val lineChunks = scrambles.map { it.splitToLineChunks(scrambleFont, scrambleColumnWidth) }
```
suushiemaniac: You moved the padding to within the `splitToLineChunks` method. Did you check that this refactor doesn't conflict with the behaviour of other PDF classes that use `splitToLineChunks` (if any)?

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/util/PdfUtil.kt
```diff
@@ -5,22 +5,20 @@ import com.itextpdf.text.Font
...
             .flatMap { it.splitLineToChunks(font, availableTextWidth) }
             .map { it.toLineWrapChunk(font) }
     }
 
-    fun String.splitLineToChunks(font: Font, availableTextWidth: Float): List<String> {
-        return splitLineToChunksRec(font, availableTextWidth, listOf())
-    }
-
-    private tailrec fun String.splitLineToChunksRec(font: Font, availableTextWidth: Float, acc: List<String>): List<String> {
+    private tailrec fun String.splitLineToChunks(font: Font, availableTextWidth: Float, acc: List<String> = listOf()): List<String> {
```
suushiemaniac: Nice catch! ;-)

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/util/PdfUtil.kt
```diff
@@ -152,22 +146,6 @@ object PdfUtil {
         }
     }
 
-    tailrec fun binarySearchInc(min: Float, max: Float, precision: Float, shouldIncrease: (Float) -> Boolean): Float {
```
suushiemaniac: I don't think this should be removed just because it is currently not used. You yourself pointed out how much troble incorrect bifurcation can cause with binary search and I'd prefer for this to "lay around unused" just in case we might ever need it.

suushiemaniac:
Concerning my remark about small font paddings:

Left = Your new version from the sample PDF in the original post
Right = TNoodle 0.15.1

2020-03-24-184222_1845x807_scrot

Note how especially towards the bottom of each table row, the scrambles in the official release have more padding before the border

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/GeneralScrambleSheet.kt
campos20: 190 was a hack introduced by someone. Half of it just sound a lot. A third sounds the correct choice. I could remove the 190 hack by splitting in new lines first.

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/GeneralScrambleSheet.kt
campos20: Nice! I kept line 89 standard though.

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/GeneralScrambleSheet.kt
campos20: Every single time split lines was called, there was a padding before. I checked and that is completely OK.

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/util/PdfUtil.kt
campos20: Ok

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/GeneralScrambleSheet.kt
campos20: I'll check on the duplication of removing stuffs. I think we were using too much constants for padding so I replaced them with 1. I'm OK with 2 (vertical and horizintal).

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/GeneralScrambleSheet.kt
campos20: This is what allowed removing megaminx hack. I'll add comments on the code.

campos20:
> Concerning my remark about small font paddings:
> 
> Left = Your new version from the sample PDF in the original post
> Right = TNoodle 0.15.1
> 
> 2020-03-24-184222_1845x807_scrot
> 
> Note how especially towards the top and bottom of each table row, the scrambles in the official release have more padding before the border

This was intentional. With a tiny border, scramble font can get a little bigger

campos20:
> What do you mean by "side by side"?

Something like this

SQ1

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/GeneralScrambleSheet.kt
campos20: Fixed

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/GeneralScrambleSheet.kt
campos20: I added a const on the companion, but 3 is just an arbitrary choice, like 2 was (and I think an imagem filling half of the width would be disturbing to the eye).

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/GeneralScrambleSheet.kt
campos20: Also, this removed the annoying M hack. There were a lot of strange things going on like `longestAlignedScrambleMaskedAndStuffed`

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/GeneralScrambleSheet.kt
campos20: I just remembered. There was a place that splitted scrambles without the padTurnsUniformly. This caused wrong calculation of the fontsize. Font size was calculated in a way (using splitlines) and when things would get rendered, splitlines worked differently. It took me forever to find, it was very frustrating. There were fireworks when I realized this.

webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/util/PdfUtil.kt
```diff
@@ -120,30 +118,26 @@ object PdfUtil {
...
-        // ideally, we could pass the object in which our text is going to be rendered
-        // as argument instead of asking leadingMultiplier, but we are currently rendering
-        // text in pdfcell, columntext and others
-        // it'd be painful to render lines in a common object to ask leadingMultiplier
+    fun fitText(font: Font, text: String, availableArea: Rectangle, maxFontSize: Float, newlinesAllowed: Boolean, leadingMultiplier: Float = 1f): Float {
         return binarySearchDec(1f, maxFontSize, FITTEXT_FONTSIZE_PRECISION) {
             // FIXME inplace modification is no good
             font.size = it
 
-            val lineChunks = text.splitToLineChunks(font, rect.width)
```
campos20: @suushiemaniac this was the place that was leaking splitToLineChunks with padTurnsUniformly. First I fixed the strange behavior that scrambles were taking more than 1 page by adding padTurnsUniformly here. Then I realized I could use inside splitToLineChunks.

campos20:
Increased scramble padding by 1 px, updated on the OP.

suushiemaniac:
I'm still seeing problems with the bottom padding, based on the PDF file that is attached to the original post at the time of this comment. (page 16)

Screenshot_20200325-171257

Also, due to the change in scramble image width, 5BLD is on 2 pages now. This is a definite No-No that *must* be avoided.

Screenshot_20200325-171308

suushiemaniac:
On a sidenote, SQ-1 sideways should not be a concern right now. It's fine the way it is until we decide to touch TNoodle-lib scramble drawing code

campos20:
> Concerning my remark about small font paddings:
> 
> Left = Your new version from the sample PDF in the original post
> Right = TNoodle 0.15.1
> 
> Note how especially towards the top and bottom of each table row, the scrambles in the official release have more padding before the border

That's completely ok to me and I'm totally fine with it. I'd like to keep small margins and we should probably get a third opinion :)


> Also, due to the change in scramble image width, 5BLD is on 2 pages now. This is a definite No-No that must be avoided.

My gosh! That slipped somehow! Fixing ASAP.

campos20:
@suushiemaniac another bit. So it's more spaced now. Updated on the OP again.

suushiemaniac:
Thanks, looking good now!

5BLD is also on 2 separate pages, but the last commit doesn't directly address the issue (ie scramble padding doesn't directly affect the page spread). Lucky coincidence or did you actually fix that as well? If so, where?

campos20:
> Thanks, looking good now!
> 
> 5BLD is also **NOT** on 2 separate pages, but the last commit doesn't directly address the issue (ie scramble padding doesn't directly affect the page spread). Lucky coincidence or did you actually fix that as well? If so, where?

I can only guess about that. My guess is that we are calculating font size wrongly somehow, so increasing padding would also mask this mistake. This wrong calculation could happen on binary search, on chunk splitting or on header/footer space.

About the header footer space, I have chosen [12 arbitrarily](https://github.com/thewca/tnoodle/blob/9be36903525b31213f78cd253a2a2dbf2a5f20e0/webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/GeneralScrambleSheet.kt#L205), but the [actual generation](https://github.com/thewca/tnoodle/blob/9be36903525b31213f78cd253a2a2dbf2a5f20e0/webscrambles/src/main/kotlin/org/worldcubeassociation/tnoodle/server/webscrambles/pdf/WatermarkPdfWrapper.kt#L65) is not aware of that.

campos20:
Typing that "out loud" made me hate that. That was on my todo list , but now it's priority 1.

campos20:
Funny fact. We are generating the PDF's default name as of GMT. I don't think this is a big deal.

PR #571, opened 2020-05-24
Force Jest runner to terminate without watch

suushiemaniac:
Yes, the double `watchAll` is intentional. See https://github.com/facebook/create-react-app/issues/6899 for details.


suushiemaniac:
WCA API seems to have choked.

PR #572, opened 2020-05-24
Add back CLI flag for manual port control

suushiemaniac:
see title.

CC @campos20


campos20:
Ws also have that proxy, but It's probably not that important.

suushiemaniac:
Yes, there is the proxy for frontend, but it is hardcoded in `package.json` and as such cannot be controlled automatically.

Also, while it may be important, it only exists for _local_ development (ie when you start the backend and frontend separately for yarn hot-reload). And as long as you're working locally, you have full control over the port binding anyways.

PR #585, opened 2020-07-04
Change UI colors fix #584

### LINKED ISSUE #584: Improve UI colo(u)rs
partial copy/paste from an email by Manu on the release testing thread:

> 1. The buttons all have very different colours (black, blue, green, blue-green-ish, white). I think a bit more consistency would make the interface more intuitive.
> 2. Related to 1, a disabled "Generate Scrambles" button is almost the same blue as the enabled one, and it's not clear that it is disabled, nor why. If I click on it, nothing happens! Only when I'm a very patient user and hover my cursor over it, without clicking, the error appears. (Of course in this case the problem should be obvious for experienced delegates, still it might be confusing for less experienced ones.)

campos20:
Preview here

https://campos20.github.io/tnoodle/

### DIFF
--- a/tnoodle-ui/src/api/tnoodle.api.js
+++ b/tnoodle-ui/src/api/tnoodle.api.js
@@ -1,5 +1,6 @@
-//let baseUrl = window.location.origin;
-let baseUrl = "https://secret-atoll-71628.herokuapp.com";
+// let baseUrl = window.location.origin;
+// let baseUrl = "http://localhost:2014";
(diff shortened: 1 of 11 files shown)


suushiemaniac:
Looks very sleek, I like the stronger contrast!
One minor remark: Is there any way to make the faded/inactive "Generate scrambles" button even more faded? I think if you use default bootstrap classes and then manually override only the alpha channel, it should be possible to make even more of a distinction between enabled/disabled

campos20:
image

We are now using the button on the left. They are side by side just on the image, not on the code

campos20:
Merging as everything was addressed.

PR #591, opened 2020-08-01
Add http interceptor

campos20:


campos20:
This is a WIP since backend should be updated as weel

suushiemaniac:
Nitpick: It seems as if all the individual `if (response.ok)` checks in `EventPickerTable` can be dropped now because the global interceptor guarantees that the response is ok anyways (otherwise an error would be thrown and execution doesn't reach that part of the code)

Not sure if this is considered good practise though

tnoodle-ui/src/App.js
```diff
@@ -2,12 +2,62 @@ import React, { Component } from "react";
...
+    handleHttpError = (error) => {
+        let message = error.statusText || JSON.stringify(error);
+        this.updateErrorMessage(message);
+    };
+
+    updateErrorMessage = (errorMessage) => {
+        // Clear the message after some seconds
+        setTimeout(() => {
+            this.setState({ ...this.state, errorMessage: "" });
+        }, 1000 * 10);
```
suushiemaniac, line 43: Magic number calculation. Better create a constant "errorTimoutInSeconds" somewhere and reference that here

tnoodle-ui/src/components/EventPickerTable.jsx
```diff
@@ -62,23 +58,12 @@ const EventPickerTable = connect(
...
         };
 
-        setFlashMessage = (text, bg, response) => {
-            this.props.updateFlashMessage(text + " " + response.statusText, bg);
-        };
-
         getFormats = () => {
             fetchFormats()
                 .then((response) => {
                     if (response.ok) {
```
suushiemaniac, line 64: Perhaps the check is not necessary anymore, cf. my global comment

tnoodle-ui/src/components/EventPickerTable.jsx
```diff
@@ -91,13 +76,6 @@ const EventPickerTable = connect(
...
```
suushiemaniac, line 76: see above (?)

tnoodle-ui/src/components/EventPickerTable.jsx
```diff
@@ -110,13 +88,6 @@ const EventPickerTable = connect(
...
```
suushiemaniac, line 88: see above (?)

tnoodle-ui/src/components/EventPickerTable.jsx
```diff
@@ -151,11 +122,6 @@ const EventPickerTable = connect(
...
```
suushiemaniac, line 122: see above (?)

tnoodle-ui/src/App.js
campos20: Updated, thanks! I just changed the "timeout" part because I think this would make it confusing somehow.

campos20:
> Nitpick: It seems as if all the individual `if (response.ok)` checks in `EventPickerTable` can be dropped now because the global interceptor guarantees that the response is ok anyways (otherwise an error would be thrown and execution doesn't reach that part of the code)
> 
> Not sure if this is considered good practise though

Well, that could cause us to try to download as zip an http message. The interceptor does not change the response or prevent it from going its regular path.

suushiemaniac:
Changes looking good! This new version resets the page entirely when clicking "Download scrambles". Could you change this back so that the state of the page remains?

campos20:
> Changes looking good! This new version resets the page entirely when clicking "Download scrambles". Could you change this back so that the state of the page remains?

Done

PR #606, opened 2020-09-06
Fix FMC translation selection in UI

suushiemaniac:
The actual fix is contained in https://github.com/thewca/tnoodle/commit/62418c948edeefe3d71cdb79a71411e35440192a

When investigating the problem, I also realised that the current reducers update translations by inverting the current `status` boolean whenever a checkbox is clicked. This may be problematic when the checkbox value and the translation status get out of sync.


campos20 (approved):

campos20:
LGTM, thanks!

PR #614, opened 2020-09-17
Hotfix for Turkish installations

suushiemaniac:
Conversion between I and i (lowercase and uppercase letter that comes between H and J) is messed up because Turkish uses an additional I without dot


suushiemaniac:
@campos20 Please review to make sure there are no typos between the UPPER_CASE enum names and the lowerCase WCIF strings

Issue #626, opened 2020-10-03
Display backend erros sent using websocket


campos20:
## Is your feature request related to a problem? Please describe.
After the introduction of websockets to display progress bars, TNoodle erros sent by the backend stopped being shown to the user. We use http interceptor to display messages.

## Describe the solution you'd like
All erros could be shown,

## Additional context
Currently, we use a component called [Interceptor](https://github.com/thewca/tnoodle/blob/master/tnoodle-ui/src/main/components/Interceptor.jsx). We could make this into a Message component, move http interceptor to some place else. Message component would display messages from both http interceptor and the websocket [error handling](https://github.com/thewca/tnoodle/blob/master/tnoodle-ui/src/main/api/tnoodle.socket.js#L23).

campos20:
Fixed by https://github.com/thewca/tnoodle/pull/628

PR #628, opened 2020-10-04
Fix websocket error handling

suushiemaniac:
The `fetch-intercept` module does not cover websockets, so I decided to process websocket errors manually and then "inject" them into the interceptor.

Feels a little bit nasty to do this with React.Ref(), is there any smarter way?


campos20:
Fixes https://github.com/thewca/tnoodle/issues/626

campos20 (approved):

PR #644, opened 2021-02-27
Update dependencies

campos20:


suushiemaniac (commented):
Why are you deleting `src/setupProxy.js`?
That file is necessary to connect to WebSockets locally, especially when using npm dev instance on port 8080 and local server on port 2014.

campos20:
> Why are you deleting `src/setupProxy.js`?
> That file is necessary to connect to WebSockets locally, especially when using npm dev instance on port 8080 and local server on port 2014.

Thanks fixing. Some other tests failed as well. Reverting some breaking changes

campos20:
Closing this. Images stopped working in some scenarios and debugging is taking too long. Waiting until we are switching to TypeScript.

PR #660, opened 2021-04-11
Migrate Interceptor.jsx to TypeScript

suushiemaniac:
Use Axios intercept mechanism as per https://github.com/axios/axios#interceptors

### DIFF
--- a/tnoodle-ui/package.json
+++ b/tnoodle-ui/package.json
@@ -14,7 +14,6 @@
         "@types/react-dom": "^17.0.0",
         "axios": "^0.21.1",
         "bootstrap": "^4.6.0",
-        "fetch-intercept": "^2.3.1",
         "lodash": "^4.17.21",
         "react": "^17.0.2",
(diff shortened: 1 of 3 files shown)


campos20:
Interceptor was there to catch error messages. I was running some [tests with antd](https://github.com/campos20/tnoodle/tree/feature/antd), it has a better message display ([choose a button](https://ant.design/components/message/) and click it), as easy as `message.info(content)`.

suushiemaniac:
Given that we don't use antd as of now, what's your opinion on merging this "as-is"?
I don't want to push this change back until you migrated the entire layout

campos20:
> Given that we don't use antd as of now, what's your opinion on merging this "as-is"?
> I don't want to push this change back until you migrated the entire layout

Please, merge it. I'm not sure about adding antd anyways. It's the second most used react-ui (after bootstrap) and we've had some contributions regarding bootstrap. Using the most used ui lib makes more sense for understaffed open-source projects. Anyways, I'll keep experimenting.

PR #718, opened 2022-03-13
Remove conflicting --watchAll flag from test script

JonEsparaz:
The test script includes the Jest CLI flag `--watchAll` _and_ `--watchAll=false`.

Experimentally, I've confirmed that the `--watchAll=false` is taking priority, so `--watchAll` has been removed.


gregorbg:
This duplication was a workaround for a bug in an older version of the JS testing framework.
When that framework got updated and the bug was fixed, we forgot to remove the manual workaround.

Thanks for spotting this!

PR #893, opened 2024-01-06
WIP Add infra for web scrambles

campos20:
Quick fix because os specific is starting to be critical.

This introduces terraform so we can run tnoodle with docker in AWS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,5 @@ gradle.properties
 
 .idea/
 .vscode/
+
+.terraform/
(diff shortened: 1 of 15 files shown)


campos20:
This is on draft until we

- [ ] Document (README)
- [ ] Deploy this to production and fix issues (we will need to create or just configure an auth client, as an example)

FinnIckler:
Does this make sense to do? I haven't heard us talk about plans like this. I thought we would migrate from tnoodle soon and as the new tool will be js based we can just run it a cdn/in the browser

gregorbg:
Yup, this was sanctioned and approved by me on quite short notice, because the recent TNoodle 1.2.1 release was basically just one giant "JVN platform is a b*tch" bug fix galore.

It's annoying to serve different operating systems with different JVMs, so Alexandre and I opted for the cloud. This will also help befriend WRC with the idea that scrambles will eventually move fully online anyways

